PR #5, opened 2015-09-22
Pull strategies and storage from bundler-api

pcarranza:
Extract what we had there into gemstash

And some fiddling to make the rubocop pass, we need to talk to that guy :disappointed:

### DIFF
--- a/.rubocop-relax.yml
+++ b/.rubocop-relax.yml
@@ -1,0 +1,11 @@
+# Offense count: 65
+# Configuration parameters: AllowURI, URISchemes.
+Metrics/LineLength:
+  Max: 122
+
+# Offense count: 85
+# Cop supports --auto-correct.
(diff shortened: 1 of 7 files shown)


.rubocop-relax.yml
```diff
@@ -0,0 +1,11 @@
+# Offense count: 65
+# Configuration parameters: AllowURI, URISchemes.
+Metrics/LineLength:
+  Max: 122
+
+# Offense count: 85
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle, SupportedStyles, UseHashRocketsWithSymbolValues.
+Style/HashSyntax:
+  Enabled: false
```
sferik: Why not autocorrect these to make the project use a consistent hash syntax?

pcarranza:
@sferik Hey, we didn't decide yet what rubocop configuration we are enabling by default, we are just adjusting that yet.

I like the idea of enabling things with auto correct by default, is way less pain than being around fixing it by hand indeed.

PR #8, opened 2015-09-26
Defer db connection test until sqlite db file exists

carpodaster:
Checking the database configuration fails because it happens before creating a custom storage dir.

#ROSSconf


smellsblue:
Thanks for the fix!

PR #10, opened 2015-09-28
Logging all the things to a file

pcarranza:
Logging works as expected, all the layers are writing in the same file.

This was tricky, puma is expecting to work with $stdout and $stderr to log "events", something where it can "write" to.

On the other hand, rack is also expecting to have something to what is can "puts" to, again $stdout/$stderr.

Every place I looked for this, whoever was speaking was saying that the way to go is just to $stdout.reopen("file"), this is ok-ish, but this does not provide any easy way to rotate the log.

After investigating how the stdlib Logger works, I realized that the object that handles the rotation of the log is Logger::LogDevice, so the way to go was to just monkey patch Logger to obtains the LogDevice, which is almost an IO, and monkey patch that other one just to provide those 2 final methods that it is looking for: puts and sync=

With this, the rotation will still happen inside the same object, so we are still using the stdlib Logger.

There is some room to improvement as depending on the source the format of the log line changes, but for now I think this is good enough


spec/gemstash/logging_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+require "tempfile"
+
+describe Gemstash::Logging do
+  before do
+    @logfile = Tempfile.create("logfile")
+    Gemstash::Logging.setup_logger(@logfile)
+  end
+  after do
+    Gemstash::Logging.reset
+    # FileUtils.remove(@logfile)
```
pcarranza: Oops, gotta recover this, will do later

lib/gemstash/logging.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    def self.raw_logger
+      return @logging_sink.raw_logger if @logging_sink
+      $stdout
+    end
+
+    def self.reset
+      @logging_sink.close if @logging_sink
+      @loggingasink = nil
```
smellsblue: I assume this is intended to be `@logging_sink = nil`

lib/gemstash/logging.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+
+      def close
+        @logger.close
+      end
+    end
+  end
+
+  #:nodoc:
+  class MyLoggerMiddleware
```
smellsblue: Where is this used? And could it be renamed? Perhaps just `LoggerMiddleware`?

lib/gemstash/logging.rb
pcarranza: Opps, right

lib/gemstash/logging.rb
pcarranza: This is left here with the intent of unifying log formats later, just not used for now.

I like that name, will rename later.

lib/gemstash/cli/start.rb
```diff
@@ -11,11 +11,17 @@ def initialize(cli)
...
-        Puma::CLI.new(args).run
+        setup_logging
+        logger = Gemstash::Logging.raw_logger
+        Puma::CLI.new(args, Puma::Events.new(logger, logger)).run
       end
 
     private
 
+      def setup_logging
+        Gemstash::Logging.setup_logger(Gemstash::Env.log_file)
```
smellsblue: I think this would be better off as
```
Gemstash::Logging.setup_logger(Gemstash::Env.base_file("server.log"))
```
and then drop `Gemstash::Env.log_file` method, along with the `:logfile` config key.

I'm worried about having too much configuration options for details that are unlikely to need to change... which grows in the amount we need to maintain down the road. Easier to strip as many config that really doesn't need to change and add them when someone has a strong use case for it (along these lines, I think I'd like to rip out some Puma config I originally put in).

lib/gemstash/logging.rb
```diff
@@ -0,0 +1,85 @@
...
+      $stdout
+    end
+
+    def self.reset
+      @logging_sink.close if @logging_sink
+      @loggingasink = nil
+    end
+
+    #:nodoc:
+    module LoggerMixin
```
smellsblue: Maybe `LoggerPatch` is a little more descriptive?

lib/gemstash/cli/start.rb
pcarranza: Agreed, makes sense.

lib/gemstash/logging.rb
pcarranza: Yes it is.

PR #11, opened 2015-10-02
Gem sources

smellsblue:
Factor gem sources to correspond to upstreams, use Rack middleware to figure out which source to use, and push all meaningful logic out of Gemstash::Web.


pcarranza:
Hold on on this change. I think there are a few things missing. Let me review better when I get home.

smellsblue:
@pcarranza thanks for the review! And no rush, I'm not able to work on gemstash till later tonight probably anyways :-)

smellsblue:
I was also thinking of adding an integration test against the redirect strategy later, I'll push to this branch when I get to it.

lib/gemstash/gem_source/upstream_source.rb
```diff
@@ -0,0 +1,143 @@
...
+      end
+
+      def serve_prerelease_specs
+        redirect web_helper.url("/prerelease_specs.4.8.gz", request.query_string)
+      end
+
+    private
+
+      def web_helper
+        @web_helper ||= Gemstash::WebHelper.new(server_url: env["gemstash.upstream"])
```
pcarranza: Just to share a thought, I think we got to the same place with 2 different paths.

The branch I am working with I am basically building the "helper" (which I renamed to HTTPClient), but I am injecting a client builder which knows how to return *something* that knows how to create a client. This is so basically because of the same thing: we are changing the upstream url.

By injecting this "builder", you can also inject it in testing, and then you don't need to stub the methods to control that in testing. But this is a WIP in my branch still.

lib/gemstash/gem_source.rb
```diff
@@ -0,0 +1,89 @@
...
+        request_uri_match = env["REQUEST_URI"].match(matcher)
+        return unless request_uri_match
+        path_info_match = env["PATH_INFO"].match(matcher)
+        return unless path_info_match
+        log_start = "Rewriting '#{env["REQUEST_URI"]}'"
+        env["REQUEST_URI"][request_uri_match.begin(0)...request_uri_match.end(0)] = ""
+        env["PATH_INFO"][path_info_match.begin(0)...path_info_match.end(0)] = ""
+        log.info "#{log_start} to '#{env["REQUEST_URI"]}'"
+        path_info_match
+      end
```
pcarranza: I think that this url parsing is complex enough to deserve an object in itself rather than a method. It is quite hard to read and there are exit conditions in the middle of the method, which makes it hard to follow.

In the branch I have in progress I actually built an upstream object that knows how to parse this url and extract the upstream URI. Which has values like `domain`, `user` and `password` which is used for gem storage. And of course the upstream URL in itself.

What do you think of pulling this logic out to a responsible object?

lib/gemstash/gem_source/rack_middleware.rb
```diff
@@ -0,0 +1,22 @@
...
+      def initialize(app)
+        @app = app
+      end
+
+      def call(env)
+        Gemstash::GemSource.sources.each do |source|
+          next unless source.matches?(env)
+          env["gemstash.gem_source"] = source
+          break
+        end
```
pcarranza: I like the idea of capturing the upstream in the middleware, that way all the methods are affected yet not changed.

lib/gemstash/gem_source.rb
```diff
@@ -0,0 +1,89 @@
...
+    autoload :UpstreamSource, "gemstash/gem_source/upstream_source"
+
+    API_REQUEST_LIMIT = 200
+
+    def self.sources
+      @sources ||= [
+        Gemstash::GemSource::PrivateSource,
+        Gemstash::GemSource::RedirectSource,
+        Gemstash::GemSource::UpstreamSource,
+        Gemstash::GemSource::RubygemsSource
```
pcarranza: I think that the only difference between an UpstreamSource and the RubygemsSource is the upstream endpoint. Does it make sense to handle that complexity inside an *upstream* object?

pcarranza:
I like it, I actually like a lot to handle the path parsing in the middleware, but I think that we are spreading the logic of how to handle the upstream (by calling chomp_path from many places) too much.

I think that the concept of the upstream as a concern is there asking to be extracted to a professional object, and that the web_helper is being stretched by not creating an HTTPClient which is what we actually need.

That said, let's move forward because I think that it is a step in the right direction. Then I can rebase my changes on top of this and start injecting the abstractions I already have in my WIP branch.

lib/gemstash/gem_source.rb
pcarranza: To be more precise, what I have is a GemRequest, and inside of it there is an upstream object. It may be usable or not. But that is something for the GemFetcher to decide.

The gem request is built with the request url.

Issue #18, opened 2015-10-06
Task to actively pre-cache all of rubygems.org


indirect:
If you're going to, say, railscamp, or mainland china, you might want to grab a copy of all the gems. Or at least the newest version of each gem. A task to make this a single command would be swell.

smellsblue:
Maybe something like `gemstash preload`? or maybe even a way to preload specific gems? like `gemstash preload rails`?

How should all versus latest work? Maybe `gemstash preload --latest`?

What if you do latest and some of them have dependencies on older versions? Should it do its best to ensure all gems have their dependencies preloaded?

pcarranza:
Where to stop?

Pull all the latest gems and their dependencies? Then cut in the first degree of the graph?

That would still be a lot, but not all the gems and versions in the world.

indirect:
no idea :) maybe just everything? I know that's what gem mirrors destined for inside the great firewall and railscamp servers do.

pcarranza:
What about a `fetch all ` plus to the `fetch latest `.

One for the great wall, the other one a bit smarter.

smellsblue:
@indirect keep in mind certain requests will still be directed towards rubygems.org, such as dependencies... those cache for 30 minutes and then will get re-fetched from rubygems.org after that... and the cache is in memory, or optionally a memcached server, so preloading won't guarantee no web requests (at least as it stands now).

Do we need to have a mode for gemstash that deals with dependencies in a different manner, or would that work?

indirect:
if gemstash caches the full index files (`specs.4.8.gz`, `latest_specs.4.8.gz`, and `prerelease_specs.4.8.gz`), then --full-index mode will work!

indirect:
it might make sense to have a flag that explicitly puts gemstash in "no backing server" mode?

pcarranza:
Offline mode, yes

pcarranza:
What about, when going in offline mode we pull the specs files, cache them and build the dependencies data without an expiration.

When we leave offline mode, we evict all this cached data.

indirect:
If you have a local server, the dependency API isn't even needed because you can fetch the full index so quickly. Plus, the new index will cache the index on each client, making the dependency API even less needed. (And it will automatically cache the index as part of the fetching process. New index support is basically Gemstash 1.1 :)

pcarranza:
Then there is no need for building the db. Just caching the specs files if enough.

And the gems, of course.

pcarranza:
[...]

PR #26, opened 2015-10-12
Avoid filesystem limits using a trie

smellsblue:
Implement storage trie based on idea from @pcarranza in discussion for #18.

All tests are passing with no additional changes... should I write a test that fails if the storage isn't using a trie though?

### DIFF
--- a/lib/gemstash/storage.rb
+++ b/lib/gemstash/storage.rb
@@ -41,11 +41,15 @@ module Gemstash
     def initialize(folder, name)
       @base_path = folder
       @name = name
-      trie_parents = @name[0...3].split("")
+      # Avoid odd characters in paths, in case of issues with the file system
+      safe_name = @name.gsub(/[^a-zA-Z0-9_]/, "_")
+      # Use a trie structure to avoid file system limits causing too many files in 1 folder
+      # Downcase to avoid issues with case insensitive file systems
+      trie_parents = safe_name[0...3].downcase.split("")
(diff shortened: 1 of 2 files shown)


pcarranza:
Let me get the weird cases we will find down the line to add tests for those and see what happens there.

smellsblue:
I can already think of a case that will fail... `-` seems fine on my machine, but I bet `.` would do bad things.

However, a simple regex gsub I think would fix it... maybe gsub anything not in a set of acceptable characters? Like:
```
trie_parents = @name[0...3].gsub(/[^0-9a-zA-Z]/, "_").split("")
```

smellsblue:
hmmm, now that I think of it, the `.` probably wouldn't really pose a problem... it would essentially get ignored in the File.join, but it wouldn't be added in a trie manner, so I think gsubbing everything outside a certain range might still be the best option

pcarranza:
This is the whole data sample of what we can find with current specs:
```
["_", "-", "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z", "."]
```

pcarranza:
So, yeah, maybe replace the . with something else that is not so full of meaning at the filesystem level.

And downcase to prevent the OSX HFS stupidity

pcarranza:
:+1:

PR #28, opened 2015-10-13
A few tweaks: enable setting the host to bind to, and reduced access for some attributes

pcarranza:
Mainly because I am running gemstash in a host on with a public IP, I don't want it to be a public mirror.

### DIFF
--- a/bin/gemstash
+++ b/bin/gemstash
@@ -1,0 +1,4 @@
+#!/usr/bin/env ruby
+
+require 'bundler/setup';
+exec 'exe/gemstash'
--- a/bin/gemstash
+++ b/bin/gemstash
@@ -1,4 +1,0 @@
-#!/usr/bin/env ruby
-
-require 'bundler/setup';
-exec 'exe/gemstash'
--- a/lib/gemstash/configuration.rb
+++ b/lib/gemstash/configuration.rb
@@ -7,8 +7,7 @@ module Gemstash
       :cache_type => "memory",
       :base_path => File.expand_path("~/.gemstash"),
       :db_adapter => "sqlite3",
-      :host => "0.0.0.0",
-      :port => 9292,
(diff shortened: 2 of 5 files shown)


pcarranza:
BTW cc @smellsblue

smellsblue:
:+1: though how about the bind have the tcp:// and port baked into the option rather than the host and port separate used to put it in?

That way, if someone wants to bind it on a unix socket and wrap it in an nginx/apache server or something, it would be pretty easy.

pcarranza:
Yip, will do it

smellsblue:
Awesome! Looks good to me! Once you merge this in, I'll update documentation branch with the changes.

pcarranza:
Ah, cool, I'll just do it.

indirect:
Awesome. We can probably just pass through `-h/--host`, `-p/--port` and `-b/--bind` directly to Puma? I know it supports all of those things. And I'm :+1: on supporting binding to unix sockets.

pcarranza:
I don't think that would be a problem, this was just attacking the configuration system.

I'll give it a spin later.

smellsblue:
I've been thinking about command line config for a while... I put my thoughts into #30. Also, the new bind option already is a pass through to Puma, should we also support host and port though? I ask because from the [Puma project](https://github.com/puma/puma#binding-tcp--sockets), it looks like the bind option is preferred over host and port.

Digging into the code a bit, the command line options for Puma doesn't actually have a host option, and [the port option](https://github.com/puma/puma/blob/master/lib/puma/cli.rb#L355) acts effectively like the bind option behind the scenes. While there is no `--host` option, I think the config DSL supports it.

Thoughts @indirect, @pcarranza?

indirect:
I must have been thinking of the config DSL. It's fine with me to only support bind and port on the command line.

Issue #27, opened 2015-10-13
Ugly trace vomited to bundler when upstream connection fails while resolving dependencies


pcarranza:
Nothing to worry about, seems to be an edge case, but it outputted a lot of data that didn't make any sense.

Probably we need to guard ourselves from failing connections.

```
[2015-10-13 15:36:10 -0400] - INFO - Fetching dependencies: netrc, mime-types, http-cookie, rdoc, simplecov-html, docile, lockfile, power_assert, ast, slop, sexp_processor, little-plugger, loquacious, bones-git, bones-extras, abstract, metaclass, needle, jruby-pageant, echoe, arel, activemodel, tzinfo, activerecord-deprecated_finders, thread_safe, minitest
[2015-10-13 15:36:15 -0400] - ERROR - 2015-10-13 15:36:15 - Faraday::ConnectionFailed - execution expired:
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/net/http.rb:879:in `initialize'
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/net/http.rb:879:in `open'
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/net/http.rb:879:in `block in connect'
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/timeout.rb:89:in `block in timeout'
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/timeout.rb:99:in `call'
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/timeout.rb:99:in `timeout'
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/net/http.rb:878:in `connect'
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/net/http.rb:863:in `do_start'
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/net/http.rb:852:in `start'
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/net/http.rb:1375:in `request'
	/home/pcarranza/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/net/http.rb:1133:in `get'
[...]

PR #29, opened 2015-10-13
Handle upstream connection error correctly and die with dignity

pcarranza:
Solving #27.

I'm just considering the case of a possible ConnectionFailed (DNS failure in my case), and in such case, retry 3 times, and then die with an error that can be interpreted by the browser rather than an error 500.

### DIFF
--- a/lib/gemstash/http_client.rb
+++ b/lib/gemstash/http_client.rb
@@ -14,7 +14,13 @@ module Gemstash
   end
 
   #:nodoc:
+  class ConnectionError < StandardError
+  end
+
+  #:nodoc:
   class HTTPClient
+    include Gemstash::Logging
+
     DEFAULT_USER_AGENT = "Gemstash/#{Gemstash::VERSION}"
 
     def self.for(upstream)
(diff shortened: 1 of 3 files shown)


lib/gemstash/http_client.rb
```diff
@@ -14,7 +14,13 @@ def initialize(message, code)
   end
 
   #:nodoc:
+  class ConnectionError < StandardError
```
smellsblue: What do you think of:
```
class ConnectionError < Gemstash::WebError
  def initialize(message)
    super(message, 502) # Bad Gateway
  end
end
```

Then the `rescue Gemstash::WebError => e` in `upstream_source.rb`  which halts with the WebError code would capture both conditions. Also, I think a `Gemstash::ConnectioError` could arguably be considered a form of a WebError.

pcarranza: Makes sense, I don't think we need to handle that case any different than the web error anyway.

I was just making the logic explicit in this first step.

smellsblue:
:+1:

PR #31, opened 2015-10-17
Add many ruby versions to travis configuration

pcarranza:
To be sure that gemstash is supported in these ruby versions at least.

cc/ @smellsblue


smellsblue:
I feel like this might be more versions than we need. I tried to find something that lists the officially supported versions, but all I could find was [the downloads page](https://www.ruby-lang.org/en/downloads/) and [a blog post about the recent Ruby versioning change](https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/).

Based on the versioning post, I think if we use `2.1.7` and `2.2.3`, and then change `2.2.3` to be updated with the latest `2.2.TEENY` version, then we are catching any API incompatible versions. We may want to also use `ruby-head` to catch anything that might take us by surprise in future Ruby versions.

smellsblue:
If we pull in #33, I'd like to see `2.0.0` added back to the list

pcarranza:
I think that supporting the whole 2.x makes sense.

Then with your change we can support older Ruby gems, but I'm not sure if we actually want to do that.

Thoughts @indirect?

smellsblue:
I'd be happy to deal with any issues that crop up for legacy RubyGems support. I think it is worth it to maintain as much of a "it works out of the box" feel as possible.

I updated #33 to isolate the legacy support to a single block of code that should be much clearer how to clean up when we want to drop support (I would say we should drop such legacy support once Ruby 2.1 is no longer supported, since earlier `TEENY` versions of it came with the problematic versions of RubyGems).

When you say whole of 2.x do you mean adding `2.0.0`, or do you mean every `TEENY` version? Based on [the ruby version policy change post](https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/), that feels like running a build against every patch release of `2.0.0` or `1.9.3`, which is why I feel like just the latest `TEENY` version would suffice.

pcarranza:
Teeny, not every patch as I understand that we are not doing anything too weird

pcarranza:
Just to make this clear. We should add then:

```
2.0
2.1
2.2
ruby-head
```

right?

smellsblue:
Yeah, but maybe `ruby-head` should be marked as `allow_failures`? That way, it won't affect our build, but we can still get an early warning of potential problems ahead.

pcarranza:
Makes sense.

smellsblue:
:+1:

PR #32, opened 2015-10-17
Add rubygems version enforcement

pcarranza:
To prevent having weird errors like `TypeError: no implicit conversion of StringIO into String` for using old versions of rubygems, we enforce the use of anything over 2.4.

There are 2 lines of defense:
* we do not allow rake to install at all if the rubygems version is older.
* we do not allow running gemstash in the same condition.

In the latter, we are also going to show how to fix the problem in the error message.

cc @indirect @smellsblue

### DIFF
--- a/lib/gemstash/cli/start.rb
+++ b/lib/gemstash/cli/start.rb
@@ -38,9 +38,10 @@ module Gemstash
       end
 
       def check_rubygems_version
-        STDERR.puts("Rubygems version is too old, please update rubygems by running: " \
-                    "gem update --system") unless
-                    Gem::Requirement.new("~> 2.4").satisfied_by?(Gem::Version.new(Gem::VERSION))
+        @cli.say(@cli.set_color("Rubygems version is too old, " \
+                                 "please update rubygems by running: " \
+                                 "gem update --system", :red)) unless
+        Gem::Requirement.new(">= 2.4").satisfied_by?(Gem::Version.new(Gem::VERSION))
       end
 
       def daemonize?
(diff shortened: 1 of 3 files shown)


pcarranza:
My bad, fixing

lib/gemstash/cli/start.rb
```diff
@@ -36,6 +37,12 @@ def store_daemonized
         Gemstash::Env.daemonized = daemonize?
       end
 
+      def check_rubygems_version
+        raise "Rubygems version is too old, please update rubygems by running: " \
+              "gem install rubygems-update && update_rubygems && gem update system" unless
```
indirect: you can update with just `gem update --system` :)

pcarranza:
Are we good to go here?

lib/gemstash/cli/start.rb
```diff
@@ -36,6 +37,12 @@ def store_daemonized
         Gemstash::Env.daemonized = daemonize?
       end
 
+      def check_rubygems_version
+        STDERR.puts("Rubygems version is too old, please update rubygems by running: " \
+                    "gem update --system") unless
+                    Gem::Requirement.new("~> 2.4").satisfied_by?(Gem::Version.new(Gem::VERSION))
```
smellsblue: How about gem requirement of ">= 2.4" instead? Otherwise this will give an erroneous `too old` message if they go to 3.0.

smellsblue: Also, do you think it makes sense to colorize the message as red or yellow to draw more attention to it, or is that overkill?

pcarranza: Yep to both, I like colors, red is so warm.

On Thu, Oct 22, 2015 at 1:25 AM, Mike Virata-Stone <notifications@github.com

smellsblue:
:+1:

PR #33, opened 2015-10-20
Older RubyGems and Ruby 2.0.0

smellsblue:
It started as an experiment based on @pcarranza's findings and #32, but I wanted to see if older RubyGems support and Ruby 2.0.0 support was reasonable to do. If we accept this, I think we should update #31 to include Ruby 2.0.0.


pcarranza:
:+1: 

I wonder if it makes sense to just use a tempfile all along, this is just for the push case, so it is quite an isolated use case already.

PR #34, opened 2015-10-23
Various fixes

smellsblue:
Faraday treats leading `/` as an absolute path, blowing away any path in the originally provided uri. I updated `Gemstash::Upstream` and `Gemstash::HTTPClient` to behave the same way, otherwise there will be bugs if you try to have an upstream that goes to something with a path.

Move headers to a nested key in the properties file of storage, leaving open the possibility to store other metadata without clashing with the headers.

Read/write content as binary, otherwise some specs can fail with differing gem contents or encoding errors.


pcarranza:
LGTM :+1:

PR #36, opened 2015-10-26
Yanking gems

### LINKED ISSUE #17: Yank a private gem
Yank private gems.

Since rubygems doesn't support yanking a host other than www.rubygems.org, we should add a temporary Thor task to make it easy.

The task should be clearly marked as temporary.

Here are my thoughts for behavior, please comment if it should differ:
* Yanked gem files will remain, in case of unyank
* Fetching gem version will result in 404
* Mark in database as yanked, and don't return in dependencies API

smellsblue:
Support yanking gems via the `gem yank` command

This fixes #17


lib/gemstash/cache.rb
```diff
@@ -11,6 +11,10 @@ def initialize(client)
       @client = client
     end
 
+    def flush
+      @client.flush
```
pcarranza: Use forwardable? not important any way.

lib/gemstash/gem_yanker.rb
```diff
@@ -0,0 +1,58 @@
...
+      invalidate_cache
+    end
+
+  private
+
+    def check_auth
+      raise Gemstash::NotAuthorizedError, "Authorization key required" if @auth_key.to_s.strip.empty?
+      auth = Authorization[@auth_key]
+      raise Gemstash::NotAuthorizedError, "Authorization key is invalid" unless auth
+      raise Gemstash::NotAuthorizedError, "Authorization key doesn't have yank access" unless auth.yank?
```
pcarranza: I find the Authorization[key] a bit strange, I think it is because it is sitting the middle of the method in between raises.

Does it make sense to turn it into that looks something like Authorization.can?(key, :yank) ?

I guess in my head it makes sense to make it symmetric with the Authorization.authorize(key, %w(yank unyank)) that we are using for the opposite use.

lib/gemstash/gem_yanker.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def update_database
+      gemstash_env.db.transaction do
+        gem_id = @db_helper.find_rubygem_id(@gem_name)
+        raise UnknownGemError, "Cannot yank an unknown gem!" unless gem_id
+        full_name = "#{@gem_name}-#{@slug}"
+        version = @db_helper.find_version_by_full_name(full_name)
+        raise UnknownVersionError, "Cannot yank an unknown version!" unless version
```
pcarranza: It seems like the search for the gem_id is only done to assert that that gem actually exists.

Given that we are talking about yanking a gem by name and version `full_name`, does it make sense to just check that we can find it by name and version?

The check for the gem_name seems to be there just for double checking, but I am not too sure how much it helps to the clarity of the method as then the gem_id is not used for anything else.

If you want to keep the check, then it can be something like

``` ruby
raise UnknownGemError ... unless @db_helper.find_rubygem_id(@gem_name)
```

BTW, does it make sense to extract the `full_name` to a method for building?

spec/gemstash/gem_yanker_spec.rb
```diff
@@ -0,0 +1,157 @@
+require "spec_helper"
+
+describe Gemstash::GemYanker do
+  let(:auth_key) { "auth-key" }
+  let(:invalid_auth_key) { "invalid-auth-key" }
+  let(:auth_key_without_permission) { "auth-key-without-permission" }
+  let(:deps) { Gemstash::Dependencies.for_private }
+  let(:gem_name) { "example" }
+  let(:gem_version) { "0.1.0" }
+  let(:gem_slug) { "#{gem_version}-ruby" }
```
pcarranza: I'm starting to have the feeling that we need a lot of things for the tests, maybe it is time to consider having some factories for this? (not for this change, just a thought)

lib/gemstash/cache.rb
smellsblue: Fixed in 195aa822fce8f413c9427f460f3aa4507a2f559c along with some other forwardable calls

lib/gemstash/gem_yanker.rb
smellsblue: The checks are identical to those in `Gemstash::GemPusher`, so I moved it all to `Authorization.check(key, permission)` in 195aa822fce8f413c9427f460f3aa4507a2f559c. I couldn't think of a way to drop the line that is bothering you in the new `check` method though without combining both error cases into 1, which I don't want to do (the error message gets passed along as the `halt` message in the `Gemstash::GemSource::PrivateSource`, which is ultimately presented to the user by the `gem` command, so I think the distinction is useful)

lib/gemstash/gem_yanker.rb
smellsblue: Addressed in 195aa822fce8f413c9427f460f3aa4507a2f559c. I do want the check for the gem name, so I used your suggestion, but in doing so I was able to drop a method, so I think it was a double win :smile:

spec/gemstash/gem_yanker_spec.rb
smellsblue: Personally, I have mixed feelings about factories, but if it's something you'd like to tackle, I'm totally fine with that.

pcarranza: I have the same mixed feelings, I just wonder if we are repeating the same object building in testing a lot.

PR #37, opened 2015-10-26
Use Sequel::Model

smellsblue:
Get rid of `Gemstash::DBHelper` and use `Sequel::Model` with corresponding model classes instead.

This is based off #36, so merging it will include both.


pcarranza:
:+1: I like it, smaller objects :smile:

PR #38, opened 2015-10-28
Build against JRuby

smellsblue:
This is mostly about running the specs (especially integration specs) against JRuby to ensure Gemstash responds with the correct gems when you are using a different platform.

However, if we want to support a JRuby version of this gem, this PR includes support for that as well... it will just be a matter of building and pushing the gem under JRuby.

A few oddities I encountered:
* Travis doesn't have full support for `JRuby 9000` yet... hence the `jruby-9.0.3.0` version in `.travis.yml`. Anything else will not install the right version of JRuby
* `Process.spawn` is broken under JRuby, I filed [a ticket against JRuby](https://github.com/jruby/jruby/issues/3428) for this, it affected us because it causes Bundler to think our Gemfile was the current working directory
* JRuby was erroring out when checking that the test servers were ready to receive requests with an error that I hadn't seen on MRI (Errno::EBADF)
* Bundler has an existing issue where there is a warning message being output when using `JRuby 9000`, I have this sectioned off in the helper so we can remove the code when Bundler gets updated
* Travis is building `JRuby 9000` in a weird way such that a separate warning message comes up (I don't see it running on my machine). I adapted the spec for this case like I did for the previous item, but I don't know when or how this might get fixed by Travis


pcarranza:
:+1:

PR #40, opened 2015-10-28
Add status command

### LINKED ISSUE #35: Add `gemstash status` command
It would be nice to be able to check if the server is running the same way you can start and stop it. :)

smellsblue:
This fixes #35


pcarranza:
:+1: go for it

Issue #43, opened 2015-10-29
DB Migration policy?


smellsblue:
I added [a migration](https://github.com/bundler/gemstash/blob/master/lib/gemstash/migrations/03_version_full_name.rb) that altered an existing table originally created in [another migration](https://github.com/bundler/gemstash/blob/master/lib/gemstash/migrations/01_gem_dependencies.rb). The migration would be simplier if it was just folded into the original one, but it would require anyone with a Gemstash server to delete their database, migrate it manually, or otherwise deal with it. I think that is likely just a few test Gemstash instances right now amongst us, but I don't know for sure.

This brings up a question about DB migration policy... should we support someone cloning the repository against master and pulling the codebase and running from there? Or do we require they strictly go through the gem? I'm personally in favor of requiring they go through the gem, otherwise we would have to be super careful about any merges into master. This would allow me to merge the 2 migrations (maybe even all migrations into 1 migration?).

Thoughts?

Should I merge the 2 migrations? All 3 migrations? Leave as is and treat each migration going forward as if there is a production database in the wild?

pcarranza:
Tricky, it really depends on how do we want to maintain the service.

My gut feeling is that we should keep each migration that is needed... once we release the first production version :smile: 

So, I am good if for now we just merge the migrations, and after releasing 1.0 we get serious about maintaining migrations as needed.

indirect:
Agreed; it seems fine to rewrite migrations until we release a 1.0, and then we should provide backwards compatibility. Too bad we don't have Rails to give us a schema.rb, then this wouldn't be a problem :P

pcarranza:
:laughing: Are you a rails evangelist @indirect ?

smellsblue:
No code needed for this, and I've already merged a few of the migrations that would benefit from it.

PR #44, opened 2015-10-31
Cache if gem is indexed in Gemstash::Storage

smellsblue:
This moves whether a gem is indexed into `Gemstash::Storage` so we don't need to do any DB queries when bundling with a `Gemfile.lock`.

This also combines a migration as described in #43.

**Important note:** There is now a small amount of file interaction inside a database transaction. I think it might be fine as it should be pretty minimal in an infrequent codepath, but it makes me uncomfortable enough to mention it here. It is nice that failed file operations would cause a DB rollback, but is this shortsighted in some way I'm not thinking of?


lib/gemstash/storage.rb
```diff
@@ -80,9 +86,9 @@ def load
...
   private
 
-    def store
+    def store(storing = :all)
       FileUtils.mkpath(@folder) unless Dir.exist?(@folder)
-      save_file(content_filename) { @content }
+      save_file(content_filename) { @content } if storing == :all
       save_file(properties_filename) { @properties.to_yaml }
       self
     end
```
pcarranza: What about?

``` ruby
def save(content, properties: nil)
  save_content(content)
  save_properties(properties) unless properties.nil?
  self
end

def update_properties(data)
  load
  save_properties(@properties.merge(data))
  self
end

def save_content(content)
  store(content_filename, content)
  @content = content
end

def save_properties(properties)
  store(properties_filename, properties.to_yam)
  @properties = properties
end

def store(filename, content)
  FileUtils.mkpath(@folder) unless Dir.exist?(@folder)
  save_file(content_filename) { content }
  self
end
```

I have problems with flags :sweat_smile:

pcarranza:
Regarding a rollback triggered by a saving failure, I don't think that it is an issue at all, I find it hard to think of a case when this will happen, and also if anyone is failing to save a yaml file probably has bigger issues than a rolled back transaction.

lib/gemstash/storage.rb
smellsblue: :+1: pushed an update

smellsblue:
@pcarranza well, the rollback triggered by saving failure is kinda what I want (otherwise either the DB or the file system will be out of sync, neither case is good, especially with `Gemstash::GemSource::PrivateSource` using the storage to determine if the gem is indexed). What I don't like is interacting with anything but the database inside a database transaction... but I think it's fine and even desirable in this case... just wondering if there might be a better way :-)

pcarranza:
I think it is fair enough. Not interacting would mean to do it with some kind of queuing system just registering the job and doing it later. Given the fact that this is such a small touch (a few bytes?) I wouldn't worry at all. It is quite an edge case.

pcarranza:
:+1:

PR #45, opened 2015-11-02
Full index bundling

### LINKED ISSUE #15: Bundle with full index against upstreams
We should be able to bundle with --full-index against upstreams (including the default server)

### LINKED ISSUE #16: Bundle with full index against private gems
We should be able to bundle with --full-index against private gems hosted in Gemstash.

smellsblue:
This fixes #15 and fixes #16.

Bundling with `--full-index` involves implementing `/private/specs.4.8.gz` and `/private/quick/Marshal.4.8/<gem>-<version>.gemspec.rz`. Upstreams already work, though perhaps we want to cache some results? Maybe that's unnecessary though.


smellsblue:
I may need to also implement `prerelease_specs` after looking at Bundler source a bit.

lib/gemstash/gem_pusher.rb
```diff
@@ -38,12 +39,27 @@ def storage
...
-      storage.resource(gem.spec.full_name).save(@content, indexed: true)
+      storage.resource(full_name).save(@content, indexed: true)
+    end
+
+    def store_gemspec
+      spec = gem.spec
+      spec = Marshal.dump(spec)
+      spec = Zlib::Deflate.deflate(spec)
+      spec_storage.resource(full_name).save(spec)
     end
```
pcarranza: Heh, are we reaching the same concerns in 2 different branches? => https://github.com/bundler/gemstash/blob/pull-all-the-gems/lib/gemstash/preload.rb#L64-L85

Should we extract something that knows how to read/write gemspecs?

lib/gemstash/gem_source/private_source.rb
```diff
@@ -67,7 +67,15 @@ def serve_info(name)
...
+        gem_full_name = id.sub(/\.gemspec\.rz\z/, "")
+        gem = storage.resource(gem_full_name)
+        halt 404 unless gem.exist?
+        gem.load
+        halt 403, "That gem has been yanked" unless gem.properties[:indexed]
+        spec = spec_storage.resource(gem_full_name)
+        halt 404 unless spec.exist?
+        content_type "application/octet-stream"
+        spec.load.content
       end
```
pcarranza: Does it make sense to extract this logic of finding the gemspec? looks like a lot of logic.

Also, what do you think of storing the content type into the gem spec properties so it is only required to forward the properties as we do with the gems?

lib/gemstash/gem_pusher.rb
smellsblue: Not quite on this particular block of code. This block is about storing the `Gem::Specification` instance so it can be retrieved when `/quick/Marshal.4.8/gemname-0.1.0.gemspec.rz` is called. There is a different class that I am using that *is* similar concern to yours, and that would be `Gemstash::SpecsBuilder` which I put in `lib/gemstash/secs_builder.rb`. The difference is that the specs.4.8.gz is a Marshalled and Gzipped array of arrays, where each array has [String<name>, Gem::Version, String<platform>], this is storing a Marshalled Zlib::Deflated instance of the Gem::Specification.

lib/gemstash/gem_source/private_source.rb
smellsblue: I'll circle back on this block after I implement prereleaase specs (since that seems to be needed as I look at Bundler code and think on it more), there is commonalities here and with the specs endpoint, so factoring the commonalities among all 3 will probably simplify this a fair amount.

Regarding the headers... that's only true for the upstream gems, where we get the headers from the upstream source. The private gem headers aren't stored in storage, and I kind of think it might be better to handle that here in the `Gemstash::GemSource::PrviateSource`.

My thinking is that the PrivateSource is all about responding to web requests for private gems, so I think it makes sense for it to know about what web headers to send. If we store the headers when the private gem is pushed... it feels a bit awkward to me to construct the headers inside `Gemstash::GemPusher`. What if we want to update which headers get sent down the road? Do we only send the updated headers for new gems at that point? Do we go back and update any private gems already pushed with the new headers? Do we check if they have the header and add it dynamically before setting the headers? All this feels more complex to me than setting the headers in the PrivateSource.

Thoughts?

lib/gemstash/gem_pusher.rb
pcarranza: Yeah, you are right, that other one seems to be a bit more common, but I still cannot point at anything that can be extracted. Let's move forward for now.

lib/gemstash/gem_source/private_source.rb
pcarranza: Yep, let's avoid the early abstraction, agreed.

Yes, you don't seem to be using the properties for anything else that is not the "indexed" and given that it is just one thing it will add little to no value, so, again, let's move on and then take a look back to see if it makes any sense to extract anything later.

lib/gemstash/gem_pusher.rb
smellsblue: :+1: I think we can take a more careful look at refactoring once both our branches are in :-)

lib/gemstash/gem_source/private_source.rb
smellsblue: :+1:

PR #47, opened 2015-11-06
Various fixes 2

### LINKED ISSUE #42: Make sure writes are thread safe
We are using `Puma` in a threaded environment. `SQLite` gest unhappy if you try to commit from 2 threads. Do we need a monitor around write uses, or does `Sequel` have a way of handling this natively? Do we need to reduce the DB connection pool to 1 when using `SQLite`?

### LINKED ISSUE #46: JRuby build is slow
The JRuby build is massively slower than MRI. I suspect this is due to JRuby warm up time happening in all the spawns in the integration specs. I've heard there are gems which can pre-warm the JVM... these may drastically speed up the build.

smellsblue:
This fixes #42 and fixes #46, along with some random things I found along the way.


spec/gemstash/storage_spec.rb
```diff
@@ -67,8 +66,8 @@
     end
 
     context "with a previously stored resource" do
-      let(:resource_id) { SecureRandom.uuid }
-      let(:content) { SecureRandom.base64 }
+      let(:resource_id) { "42" }
+      let(:content) { "zapatito" }
```
pcarranza: :laughing: 

42 is a perfect randomly generated number!

spec/support/jruby_binstubs/bundle
```diff
@@ -0,0 +1,25 @@
...
+module Bundler
+  #:nodoc:
+  class Fetcher
+    # The Bundler user_agent uses SecureRandom, which causes the specs
+    # to run out of entropy and run a lot longer than they need to.
+    def user_agent
+      "gemstash spec"
+    end
+  end
+end
```
pcarranza: I wonder if it makes sense to create a Random generator based on a SecureRandom seed and just keep it around.

No a blocker, though

pcarranza:
Looks good, go for it :+1:

spec/support/jruby_binstubs/bundle
smellsblue: I don't think it's necessary, this is only in use for JRuby to speed up the build there. MRI doesn't seem to have the same problem, and the random number being generated doesn't really have any benefit in our specs. André pointed out in chat, it's for matching requests, which we don't really care about in the specs (nothing is hitting the actual rubygems servers).

PR #48, opened 2015-11-06
Pull all the gems

### LINKED ISSUE #18: Task to actively pre-cache all of rubygems.org

pcarranza:
This fixes #18 by adding a thor client to run this process.

Check the reference documentation for details on how to use it.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -313,7 +313,11 @@ you will need to pass the `--config-file` option to all Gemstash commands.
 
 ## Preload
 
-Preloads all the gems in your gemstash server from the default upstream
+Preload all the gems in your gemstash server from the default upstream.
+This can be useful if you plan to go somewhere where the internet connection
+will not be reliable or available.
+It is worth noting that this will require a large amount of disk space and
+computing time.
 
 ### Usage
 
@@ -332,35 +336,45 @@ gemstash preload --threads=10 --server_url=http://my-server-url:9292
 **Usage:** `--latest`
 
 **Description**<br />
-Only fetch the latest specs
+Only fetch the latest specs. This will limit the amount of disk space required,
+but there is no guarantee that the gems you are using will have all their
+dependencies met.
 
 #### --skip
 
 **Usage:** `--skip=<number>`
 
 **Description**<br />
-Number of gems to skip when preloading
+Number of gems to skip when preloading. This can be useful to load in chunks or
+to resume a gem preloading.
 
 #### --limit
 
 **Usage:** `--limit=<number>`
 
 **Description**<br />
-Limit for the number of gems to preload
+Limit for the number of gems to preload. In conjunction with [--skip](#--skip)
+this can be used for loading gems in chunks or resuming the process.
 
 #### --threads
 
 **Usage:** `--threads=<number>`
 
 **Description**<br />
-Number of threads to run the fetching job with
+Number of threads to run the fetching job with.
+Since this activity is heavily IO bound (most of the time goes into waiting the
+network) it is useful to use more threads for performing the task. The default
+number of threads is 20 already which should be good enough, but in case you
+want to experiment this value can be changed.
 
 #### --server_url
 
 **Usage:** `--server_url=<url>`
 
 **Description**<br />
-Gemstash server url
+Gemstash server url. By default this is assumed to be running from the same
+host as the gemstash server so it will default to `http://localhost:9292` or
+whatever is defined in your configuration as the *bind* address.
 
 ---
 
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -37,6 +37,14 @@ Table of Contents
         * [--redo](#--redo)
         * [--debug](#--debug)
         * [--config-file](#--config-file-4)
+    * [Preload](#preload)
+      * [Usage](#usage-5)
+      * [Options](#options-5)
+        * [--latest](#--latest)
+        * [--skip](#--skip)
+        * [--limit](#--limit)
+        * [--threads](#--threads)
+        * [--server_url](#--server_url)
 
 
 
@@ -302,6 +310,57 @@ This will do nothing if all checks pass.
 Specify the config file to write to. Without this option, your configuration
 will be written to `~/.gemstash/config.yml`. If you write to a custom location,
 you will need to pass the `--config-file` option to all Gemstash commands.
+
+## Preload
+
+Preloads all the gems in your gemstash server from the default upstream
+
+### Usage
+
+```
+gemstash preload
+gemstash preload --latest
+gemstash preload --limit=1000
+gemstash preload --limit=1000 --skip=1000
+gemstash preload --threads=10 --server_url=http://my-server-url:9292
+```
+
+### Options
+
+#### --latest
+
+**Usage:** `--latest`
+
+**Description**<br />
+Only fetch the latest specs
+
+#### --skip
+
+**Usage:** `--skip=<number>`
+
+**Description**<br />
+Number of gems to skip when preloading
+
+#### --limit
+
+**Usage:** `--limit=<number>`
+
+**Description**<br />
+Limit for the number of gems to preload
+
+#### --threads
+
+**Usage:** `--threads=<number>`
+
+**Description**<br />
+Number of threads to run the fetching job with
+
+#### --server_url
+
+**Usage:** `--server_url=<url>`
+
+**Description**<br />
+Gemstash server url
 
 ---
 
--- a/lib/gemstash/cli.rb
+++ b/lib/gemstash/cli.rb
@@ -10,6 +10,7 @@ module Gemstash
     autoload :Start,     "gemstash/cli/start"
     autoload :Status,    "gemstash/cli/status"
     autoload :Stop,      "gemstash/cli/stop"
+    autoload :Preload,   "gemstash/cli/preload"
 
     # Thor::Error for the CLI, which colors the message red.
     class Error < Thor::Error
@@ -65,6 +66,17 @@ module Gemstash
       "Config file to load when stopping"
     def stop
       Gemstash::CLI::Stop.new(self).run
+    end
+
+    desc "preload", "Preloads all the gems in your gemstash server from the default upstream"
+    method_option :latest, :type => :boolean, :default => false, :desc =>
+      "Only fetch the latest specs"
+    method_option :server_url, :type => :string, :default => "http://localhost:9292", :desc =>
+      "Gemstash server url"
+    method_option :threads, :type => :numeric, :default => 20, :desc =>
+      "Number of threads to run the fetching job with"
+    def preload
+      Gemstash::CLI::Preload.new(self).run
     end
   end
 end
--- a/lib/gemstash/cli.rb
+++ b/lib/gemstash/cli.rb
@@ -75,6 +75,10 @@ module Gemstash
       "Gemstash server url"
     method_option :threads, :type => :numeric, :default => 20, :desc =>
       "Number of threads to run the fetching job with"
+    method_option :limit, :type => :numeric, :default => nil, :desc =>
+      "Limit for the number of gems to preload"
+    method_option :skip, :type => :numeric, :default => 0, :desc =>
+      "Number of gems to skip when preloading"
     def preload
       Gemstash::CLI::Preload.new(self).run
     end
--- a/lib/gemstash/cli/preload.rb
+++ b/lib/gemstash/cli/preload.rb
@@ -13,10 +13,17 @@ module Gemstash
       end
 
       def run
-        latest = @cli.options[:latest]
-        threads = @cli.options[:threads]
+        return unless are_you_sure?
         http_client = HTTPClient.for(Upstream.new(@cli.options[:server_url]))
-        Gemstash::Preload::GemPreloader.new(http_client, latest, threads)
+        preloader = Gemstash::Preload::GemPreloader.new(http_client, latest: @cli.options[:latest]).
+          limit(@cli.options[:limit]).skip(@cli.options[:skip]).threads(@cli.options[:threads])
+        preloader.preload
+      end
+
+      def are_you_sure?
+        @cli.say @cli.set_color("Preloading all the gems is an extremely heavy and long running process", :yellow)
+        @cli.say @cli.set_color("You can expect that this will take around 24hs and use over 100G of disk", :yellow)
+        @cli.yes? "Are you sure you want to do this?"
       end
     end
   end
--- a/lib/gemstash/cli/preload.rb
+++ b/lib/gemstash/cli/preload.rb
@@ -1,0 +1,19 @@
+require "gemstash"
+
+module Gemstash
+  class CLI
+    # This implements the command line preload task to cache all the available gems:
+    # $ gemstash preload
+    class Preload
+      include Gemstash::Env::Helper
+
+      def initialize(cli)
+        Gemstash::Env.current = Gemstash::Env.new
+        @cli = cli
+      end
+
+      def run
+      end
+    end
+  end
+end
--- a/lib/gemstash/cli/preload.rb
+++ b/lib/gemstash/cli/preload.rb
@@ -13,6 +13,10 @@ module Gemstash
       end
 
       def run
+        latest = @cli.options[:latest]
+        threads = @cli.options[:threads]
+        http_client = HTTPClient.for(Upstream.new(@cli.options[:server_url]))
+        Gemstash::Preload::GemPreloader.new(http_client, latest, threads)
       end
     end
   end
--- a/lib/gemstash/cli/preload.rb
+++ b/lib/gemstash/cli/preload.rb
@@ -18,6 +18,7 @@ module Gemstash
         preloader = Gemstash::Preload::GemPreloader.new(http_client, latest: @cli.options[:latest]).
           limit(@cli.options[:limit]).skip(@cli.options[:skip]).threads(@cli.options[:threads])
         preloader.preload
+        @cli.say "\nDone"
       end
 
       def are_you_sure?
--- a/lib/gemstash/http_client.rb
+++ b/lib/gemstash/http_client.rb
@@ -43,15 +43,7 @@ module Gemstash
     end
 
     def get(path)
-      response = with_retries do
-        @client.get(path) do |req|
-          req.headers["User-Agent"] = @user_agent
-          req.options.open_timeout = 2
-        end
-      end
-
-      raise Gemstash::WebError.new(response.body, response.status) unless response.success?
-
+      response = request(:get, path)
       if block_given?
         yield(response.body, response.headers)
       else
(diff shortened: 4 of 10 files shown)


lib/gemstash/cli.rb
```diff
@@ -66,5 +67,20 @@ def status
     def stop
       Gemstash::CLI::Stop.new(self).run
     end
+
+    desc "preload", "Preloads all the gems in your gemstash server from the default upstream"
+    method_option :latest, :type => :boolean, :default => false, :desc =>
+      "Only fetch the latest specs"
+    method_option :server_url, :type => :string, :default => "http://localhost:9292", :desc =>
+      "Gemstash server url"
```
smellsblue: If I'm reading the code right, it looks like this is the upstream URL, not the Gemstash URL. If that's the case, that seems like it conflicts with the task description discussing the `default upstream`. If this is the upstream, I think it should probably be `nil` by default, so the task description will be correctly pulling from the `default upstream` when this is `nil` (which may require some tweaks to the code).

lib/gemstash/cli.rb
```diff
@@ -66,5 +67,20 @@ def status
...
+    method_option :threads, :type => :numeric, :default => 20, :desc =>
+      "Number of threads to run the fetching job with"
+    method_option :limit, :type => :numeric, :default => nil, :desc =>
+      "Limit for the number of gems to preload"
+    method_option :skip, :type => :numeric, :default => 0, :desc =>
+      "Number of gems to skip when preloading"
+    def preload
```
smellsblue: This task should probably support the `--config-file` option like in the other tasks. Perhaps it is time for a base CLI task class of some kind to put common behavior? Maybe we can make `--config-file` a class level option? (it's been a while since I've used it, but I believe Thor supports that)

lib/gemstash/cli/preload.rb
```diff
@@ -0,0 +1,31 @@
...
+      def initialize(cli)
+        Gemstash::Env.current = Gemstash::Env.new
+        @cli = cli
+      end
+
+      def run
+        return unless are_you_sure?
+        http_client = HTTPClient.for(Upstream.new(@cli.options[:server_url]))
+        preloader = Gemstash::Preload::GemPreloader.new(http_client, latest: @cli.options[:latest]).
+          limit(@cli.options[:limit]).skip(@cli.options[:skip]).threads(@cli.options[:threads])
```
smellsblue: Would it make sense to pass all of `@cli.options` as the last argument to the constructor, and let the constructor set all of the values based on that?

lib/gemstash/cli.rb
pcarranza: No, it is not the upstream url, I do create an `Upstream` just to reuse the object with the client (https://github.com/bundler/gemstash/pull/48/files#diff-b6f92f551dd149bd7013ca6f7f0fd7a2R17)

What this does is create an HTTPClient and start requesting the *head* of the gems to gemstash without using any upstream path, so *it* downloads those from the default upstream. This process will not make any call to rubygems.org directly.

All this does is fake a client for every gem version there is in the specs file.

lib/gemstash/http_client.rb
```diff
@@ -59,8 +51,28 @@ def get(path)
...
   private
 
+    def request(method, path)
+      response = with_retries do
+        @client.public_send(method, path) do |request|
+          request.headers["User-Agent"] = @user_agent
+          request.options.open_timeout = 2
+        end
+      end
+      raise Gemstash::WebError.new(response.body, response.status) unless response.success?
```
smellsblue: With a head request, does the body come back empty or null? If so, should we provide some sensible default message when that is the case? Keep in mind that the message is likely to be displayed to the user though, so if it is somehow useful to them, that would be extra helpful

lib/gemstash/cli.rb
pcarranza: The --config-file makes sense, but I also like to think about this as a client of gemstash. It is bundled there, but it is a one off and is 'kinda' out of the regular execution.

What would be the gain here? having the bind url? unless the server is remote, in which case we would need a specific config file for just one value, unless we build a config file for just this process which seems to be overkill

lib/gemstash/cli/preload.rb
pcarranza: Sure, will give it a try, thanks!

docs/reference.md
```diff
@@ -303,6 +311,71 @@ Specify the config file to write to. Without this option, your configuration
...
+**Usage:** `--threads=<number>`
+
+**Description**<br />
+Number of threads to run the fetching job with.
+Since this activity is heavily IO bound (most of the time goes into waiting the
+network) it is useful to use more threads for performing the task. The default
+number of threads is 20 already which should be good enough, but in case you
+want to experiment this value can be changed.
+
+#### --server_url
```
smellsblue: I think Thor converts underscores to dashes, so it would be `--server-url` (you can check by running `gemstash help preload`).

pcarranza: Ah, cool, didn't knew this.

lib/gemstash/http_client.rb
pcarranza: The head is so gemstash does not push the body to this client as it is not necessary, but gemstash will make a full get to the upstream, so it will cache the gem. But will not write the body back to this client.

lib/gemstash/cli.rb
smellsblue: Ah, gotchya

lib/gemstash/cli.rb
smellsblue: Hmmm, ok, I misunderstood this task a bit then. I fear this might be a bit confusing for a user... every other command on gemstash is oriented as a server side action, so this feels a bit out of place as a client action here (to me). Should we consider a separate client program, like what Stickler does? Or maybe re-orient the actions with a prefix... like `gemstash server ...` or perhaps only for client actions like `gemstash client preload ...`? Maybe that's excessive if this ends up as the only client side Gemstash action.

Thoughts?

docs/reference.md
```diff
@@ -303,6 +311,71 @@ Specify the config file to write to. Without this option, your configuration
...
+Number of gems to skip when preloading. This can be useful to load in chunks or
+to resume a gem preloading.
+
+#### --limit
+
+**Usage:** `--limit=<number>`
+
+**Description**<br />
+Limit for the number of gems to preload. In conjunction with [--skip](#--skip)
+this can be used for loading gems in chunks or resuming the process.
```
smellsblue: When I was inspecting the `specs.4.8.gz` from rubygems.org, it looked like all the specs were coming back alphabetically. As new gems are pushed and yanked, it seems that might cause issues with using limit and skip? I suppose if only pushes are happening, it would only cause some gems to be handled duplicate times, but if yanks happen, you could potentially miss some gem versions. Should we worry about this in some way? Should we just shrug and make a note in the limit and skip docs that you might miss gems if a yank happens with just the right timing (probably in the config CLI description too if so)?

lib/gemstash/preload.rb
```diff
@@ -0,0 +1,127 @@
...
+          end
+        end
+        pool.shutdown
+      end
+
+    private
+
+      def each_gem
+        gem_specs = @specs.fetch.to_a
+        return if !@limit.nil? && @limit <= 0
```
smellsblue: This could be `@limit && @limit <= 0`

lib/gemstash/preload.rb
```diff
@@ -0,0 +1,127 @@
...
+        return if @skip >= gem_specs.size
+        gem_specs[range_for(gem_specs)].each_with_index do |gem, index|
+          yield gem.to_s, index + @skip + 1, gem_specs.size
+        end
+      end
+
+      def range_for(gem_specs)
+        limit = (@limit || gem_specs.size) + @skip - 1
+        (@skip..limit)
```
smellsblue: I think you can drop the `- 1` if you use 3 dots here

docs/reference.md
```diff
@@ -303,6 +311,71 @@ Specify the config file to write to. Without this option, your configuration
...
+gemstash preload
+gemstash preload --latest
+gemstash preload --limit=1000
+gemstash preload --limit=1000 --skip=1000
+gemstash preload --threads=10 --server_url=http://my-server-url:9292
+```
+
+### Options
+
+#### --latest
```
smellsblue: What about --prerelease? Might want to support that as well...?

lib/gemstash/cli.rb
pcarranza: I wonder the same thing, on the one hand it makes sense to be shipped with gemstash as I like it to be a gemstash *feature* (another thing is how it manages to provide the feature), on the other hand it behaves as a batch client.

Maybe gemstash batch? and then we can use that batch for things like import gems in a batch and those other things that we talked about at some point?

docs/reference.md
pcarranza: This will download the full specs and go through that list. There is a chance that if you stop the process and restart you may skip less gems if another gem was added, because as you say the gems are sorted alphabetically. But I don't think we should worry too much about it.

One thing I was thinking that may be a cool feature is to provide a flag to *match* gem names and maybe versions, so we could fetch all the *rails* or *active.+* in one go.

But I wanted to keep this simple for now and just push the feature up.

lib/gemstash/preload.rb
pcarranza: Yep, cool, will do.

lib/gemstash/preload.rb
pcarranza: Ah, cool, didn't knew about this one.

docs/reference.md
pcarranza: Good question, don't know, makes sense to do it? it is not a big deal at all.

docs/reference.md
smellsblue: Here's the scenario I'm thinking of that might be most potential for problems:

Lets say I have iterated over 10,000 gems. Let's say gem 10,001 happens to be an important one for me. Someone yanks a gem in the first 10,000, say gem 100. Gem 10,001 is now 10,000 in specs.4.8.gz, so when I run the next block, I start on what would have been gem 10,002.

Granted, only the gems on the borders of your limits are at risk, but given how long this takes, a yank becomes more likely. It also might not be a big problem most of the time, so maybe it's not worth worrying about.

lib/gemstash/cli.rb
smellsblue: Yeah, but those could all be server side only, and probably make more sense that way anyways (importing a batch locally would probably be faster if you have the gems locally already, and then we can skip worrying about authorization keys).

Also, we have a lot of gem requirements that you really don't need if you are running client side stuff, so maybe a separate gem would make sense. However, that adds complexity, and the simplicity of just 1 gem is nice too.

Maybe we should get some advice from André?

lib/gemstash/preload.rb
```diff
@@ -0,0 +1,127 @@
...
+        @specs_file = "specs.4.8.gz" unless latest
+        @specs_file ||= "latest_specs.4.8.gz"
+      end
+
+      def fetch
+        reader = Zlib::GzipReader.new(
+          StringIO.new(@http_client.get(@specs_file)))
+        @specs = Marshal.load(reader.read)
+        self
+      end
```
smellsblue: Is there a case where you don't want to fetch the specs? Maybe this should be a private method called by the constructor?

I don't know if it really matters since this is using a `StringIO` but could you add an ensure block with `reader.close if reader`? Who knows what GzipReader is really doing... it might be opening handles and whatnot that should be closed to avoid resource leaks.

lib/gemstash/preload.rb
```diff
@@ -0,0 +1,127 @@
...
+
+      def each(&block)
+        @specs.each do |gem|
+          yield GemName.new(gem)
```
smellsblue: It looks like you are including `Enumerable` and implementing `each` and then just calling `.to_a`. Perhaps a method that maps (and caches to instance variable) @specs to `GemName` instances might be more effective? It would also drastically reduce the exposed API of this class, which I think is a benefit if it doesn't really need to be an enumerable.

If you keep it as is though, you can drop `&block` since you use `yield`

lib/gemstash/preload.rb
```diff
@@ -0,0 +1,127 @@
...
+
+      def schedule(*args, &block)
+        @jobs << [block, args]
+      end
+
+      def shutdown
+        @size.times do
+          schedule { throw :exit }
+        end
+        Thread.pass until @jobs.empty?
```
smellsblue: Did you consider pulling in Celluloid or some other 3rd party tool to handle the thread pool? It seems Celluloid has [built in pools](https://github.com/celluloid/celluloid/wiki/Pools).

If you don't go for a 3rd party implementation, I would recommend joining on the threads rather than passing, otherwise it seems like wasted cycles to let Ruby allow this thread to do work.

Also it seems like if one of the jobs triggers an exception, the thread will die... if enough failures happened, you could end up with no more threads, and the process will just hang. All the more reason to consider letting another gem do the job :-)

spec/gemstash/preload_spec.rb
```diff
@@ -0,0 +1,93 @@
+require "spec_helper"
+
+describe Gemstash::Preload do
+  let(:stubs) { Faraday::Adapter::Test::Stubs.new }
+  let(:http_client) { Gemstash::HTTPClient.new(Faraday.new {|builder| builder.adapter(:test, stubs) }) }
+  let(:latest_specs) do
+    to_marshaled_gzipped_bytes([["latest_gem", "1.0.0", ""]])
+  end
+  let(:full_specs) do
+    to_marshaled_gzipped_bytes([["latest_gem", "1.0.0", ""], ["other", "0.1.0", ""]])
```
smellsblue: The platform seems to be "ruby" rather than an empty string when the specs are fetched from rubygems.org.

Also, I just pulled latest_specs and inspected the result... it appears that it is all gems indicating the latest version, not anything to do with when gems are pushed. Thus, latest would have both latest_gem and other, 1 version each.... full_specs on the other hand would have both, but all (non prerelease) versions of the gems.

spec/gemstash/preload_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      expect(out.string).to eq("\r1/2\r2/2")
+    end
+  end
+
+  def to_marshaled_gzipped_bytes(obj)
+    buffer = StringIO.new
+    gzip = Zlib::GzipWriter.new(buffer)
+    gzip.write(Marshal.dump(obj))
+    gzip.close
+    buffer.string
```
smellsblue: See `spec/support/file_helpers.rb`, you can change this to simply `gzip(Marshal.dump(obj))`, which might be worth inlining rather than having a method for it (or at least, that's how I've done it in my specs).

lib/gemstash/cli.rb
pcarranza: Yeah, it is tricky.

What I like about the client approach is that we can exercise the API and not have special cases for things like batch import or this other batch preload case. That simplifies the server as we do not have to branch the code depending where the call comes from.

On the other hand, it does have specific needs (like celluloid if we ever want to us it), but I do not see much more needs added, at least in this particular case, to justify extracting a gem to not affect gemstash with client specific dependencies.

What do you think about this @indirect ?

docs/reference.md
pcarranza: I think I wouldn't worry about this. If you ever get into the case of "I want to fetch all the gems" you would probably just run agains all the gems and let it be.

If you need to make a second run, you have to consider that gem stash will not make a new http request to ruby gems as the gems are already stored locally in gem stash, so those calls will only go as far as loading the properties file, which is not a huge impact.

So I don't think we need to worry too much about this, at least not for now.

Still I will run some tests to see how much of a difference it is to load, say, 50,000 gems and then retry again. Just for curiosity.

lib/gemstash/preload.rb
pcarranza: I kinda do not like to have side effects in constructors if I can avoid it. 

Granted that it may make sense to make it private and that I can ensure to close the reader to no leave a resource open.

I'll just make sure that it is closed in the end.

lib/gemstash/preload.rb
pcarranza: True and true, will simplify. Thanks

lib/gemstash/preload.rb
pcarranza: Yeah, didn't wanted to add a dependency to not mess up, but did not payed too much attention to the cleanup actually.

Will give it a spin and depending on if we remove this from here or we keep it consider adding celluloid or just let it be.

spec/gemstash/preload_spec.rb
pcarranza: Yeah, the tests are more in the style of one vs two, the naming is just to cover the use of one or the other depending on the expectation, not checking multiple versions of the same gem as I do not think that is an issue to be handled here.

Also, the third value is empty basically because we are ignoring it when fetching the gem (we just build the url by gem name and version) so it is there just to comply with the expectation of what will be read from the specs file.

spec/gemstash/preload_spec.rb
pcarranza: Cool, I'll check it out.

lib/gemstash/cli/preload.rb
pcarranza: ah, no, I am actually using the client to say and ask things. I need it for that.

lib/gemstash/preload.rb
pcarranza: There are a few methods that are being called, mostly from specs, but let's take a look at how it looks like.

lib/gemstash/cli/preload.rb
smellsblue: I meant the constructor of `Gemstash::Preload::GemPreloader`, so something like this:
```
preloader = Gemstash::Preload::GemPreloader.new(http_client, @cli.options)
```

And then the `Gemstash::Preload::GemPreloader` would store all the `:latest`, `:limit`, `:skip`, and `:threads` options directly rather than using additional builder methods (or perhaps store the options object directly and use them as needed from there).

pcarranza: Gotcha, yeah that makes sense.

lib/gemstash/cli.rb
pcarranza: Ok, let's think it this way:

What about supporting initially this **client** feature (even though I plan to use the GemFetcher object in the immediate future) and just open the door to support preloading a Gemfile.lock file?

Being honest, that case will probably be the one that makes the most sense by warming up gemstash with the versions you need for the projects you are actually working on.

How that sounds?

lib/gemstash/cli/preload.rb
pcarranza: Done

docs/reference.md
pcarranza: Done

spec/gemstash/preload_spec.rb
pcarranza: Done

pcarranza:
Thanks for the feedback.

Next step, I'll give it a spin to use the `parallel` gem instead of homebrewed thread pool and change the head method for GemFetcher direct call, will see how it works.

pcarranza:
At this point I'm still using head to download the gems.

What would be left now is extracting the gem fetching and storing out of the upstream source so I can build one and not use the server. As is it is capable of fetching all the gems (all, prerelease and latest) just that it requires gemstash to be there to hit it.

docs/reference.md
```diff
@@ -313,11 +322,86 @@ Show what version of Gemstash you are using.
 ### Usage
 
 ```
+<<<<<<< HEAD
```
olleolleolle: (Um, is this a Git conflict marker?)

PR #52, opened 2015-11-29
Support mysql as DB backend

### LINKED ISSUE #51: Add support for mysql as DB backend
Just that.

pcarranza:
This provides mysql as DB backend, and also reduces the size of the authorization key because mysql does not support keys so large (and they are not necessary)

Fixes #51 

cc @indirect @smellsblue

### DIFF
--- a/lib/gemstash/cli/setup.rb
+++ b/lib/gemstash/cli/setup.rb
@@ -77,27 +77,27 @@ module Gemstash
 
       def ask_database
         say_current_config(:db_adapter, "Current database adapter")
-        options = %w(sqlite3 postgres)
+        options = %w(sqlite3 postgres mysql)
         database = nil
 
         until database
-          database = @cli.ask "What database adapter? [SQLITE3, postgres]"
+          database = @cli.ask "What database adapter? [SQLITE3, postgres, mysql]"
           database = database.downcase
           database = "sqlite3" if database.empty?
           database = nil unless options.include?(database)
         end
 
         @config[:db_adapter] = database
-        ask_postgres_details if database == "postgres"
+        ask_database_details(database) unless database == "sqlite3"
       end
 
-      def ask_postgres_details
+      def ask_database_details(database = "postgres")
         say_current_config(:db_url, "Current database url")
 
         if RUBY_PLATFORM == "java"
-          default_value = "jdbc:postgres:///gemstash"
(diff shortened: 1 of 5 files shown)


lib/gemstash/cli/setup.rb
```diff
@@ -77,27 +77,27 @@ def ask_memcached_details
...
           database = nil unless options.include?(database)
         end
 
         @config[:db_adapter] = database
-        ask_postgres_details if database == "postgres"
+        ask_database_details(database) unless database == "sqlite3"
       end
 
-      def ask_postgres_details
+      def ask_database_details(database = "postgres")
```
smellsblue: Is the default value necessary? It seems like a value is always provided

lib/gemstash/migrations/02_authorizations.rb
```diff
@@ -2,7 +2,7 @@
   change do
     create_table :authorizations do
       primary_key :id
-      String :auth_key, :size => 2056, :null => false
+      String :auth_key, :size => 255, :null => false
```
smellsblue: I believe the auth key can be user provided. The generated value is used if the user doesn't explicitly give a value. Should this limit be indicated in the docs perhaps? Is it possible there will come a day when 255 won't be enough for a secure auth token? Should we consider using a data type that works with the larger size with MySQL rather than reduce the size? Or should we document the limit and just migrate the table later if we want to expand it?

pcarranza: Docs make sense.

From the enoughness point of view, 32 chars (+4 dashes) is the length of a uuid. UUIDs are considered as universal unique ids. There is a really [good page on the wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier#Random_UUID_probability_of_duplicates) that explains what is the probability of finding a duplicate (a thing that can actually happen) and it looks to be highly unlikely.

In other words, I wouldn't worry about running out of uuids any time soon, and we still have room for putting 7 uuids in a row as an auth key.

lib/gemstash/cli/setup.rb
pcarranza: Not really, it was just left there in the process of refactoring this as a safety measure.

I'll drop it.

smellsblue:
:+1:

PR #55, opened 2015-12-04
Use ternary operator for one-line condition

farukaydin:
Although it's not a big deal, I recommend using ternary operator for one-line conditions.

### DIFF
--- a/lib/gemstash/cli/start.rb
+++ b/lib/gemstash/cli/start.rb
@@ -41,11 +41,7 @@ module Gemstash
       end
 
       def daemonize_args
-        if daemonize?
-          ["--daemon"]
-        else
-          []
-        end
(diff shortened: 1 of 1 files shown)


smellsblue:
Personally, I'm not a fan of the ternary operator... seen it abused to the point where I avoid it. I realize I'm probably a strong minority on that, though, so if it passes Rubocop and @pcarranza is happy with it, I'm cool with this.

pcarranza:
I'm in the same page as @smellsblue, not a fan of the `iif` form as it is too dense and forces me to stop and think too much about that one line => it is at least 3 lines condensed in one.

I do not see a reason to merge it. It is just a style change.

farukaydin:
Fair enough, I'm closing it since both of you recommend. Thanks for review.

Issue #56, opened 2015-12-04
Gemstash command outputs


farukaydin:
Currently, ```gemstash start``` doesn't return any output, and ```gemstash stop``` returns ```Command stop sent success```.

I think it's better to return outputs like "Gemstash started/stopped successfully!". If you're :+1: on this, I can work on it.

smellsblue:
Aboslutely, thanks! The stop output is coming from Puma, I think it's probably ok to leave that output and add the Gemstash output, unless it's trivial to block it.

PR #58, opened 2015-12-06
Add start/stop command gemstash outputs

farukaydin:
Fixes https://github.com/bundler/gemstash/issues/56

I didn't add any test since I think it doesn't need.

### DIFF
--- a/lib/gemstash/cli/start.rb
+++ b/lib/gemstash/cli/start.rb
@@ -10,6 +10,7 @@ module Gemstash
         prepare
         setup_logging
         store_daemonized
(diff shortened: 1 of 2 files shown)


pcarranza:
Just wonder if it makes more sense to do it with `cli.say` rather than the logger. Then you can even say it in green :smile: 

@smellsblue thoughts?

farukaydin:
@pcarranza updated like you mentioned. I think it's better (and green :smile:)

pcarranza:
LFTM :+1: 

Let's wait for @smellsblue if you don't mind

smellsblue:
:+1:

PR #59, opened 2015-12-06
Fix docs 'gemstash' command as 'gemstash authorize' for api key

farukaydin:
Use "gemstash authorize" command instead of "gemstash" for private gems doc.


pcarranza:
:+1:

farukaydin:
:rocket: :tada:

PR #64, opened 2015-12-14
utf8 is 4 bits on MySQL >= 5.5

chriseckhardt:
This works around the error 'Mysql::Error: Index column size too large. The
maximum column size is 767 bytes'


pcarranza:
Awesome! thanks!

Let's wait for the checks and I'll just merge it in :+1:

PR #68, opened 2015-12-23
Index cached gems

smellsblue:
Store upstream and gem name of cached gems in the database, which will allow plugins to know what gems have been cached, and easily access them.

This also fixes a bug where headers from the spec would override the headers from the gem and vice versa.

### DIFF
--- a/lib/gemstash/gem_source/upstream_source.rb
+++ b/lib/gemstash/gem_source/upstream_source.rb
@@ -160,7 +160,13 @@ module Gemstash
       def fetch_remote_gem(gem_name, gem_resource, resource_type)
         log.info "Gem #{gem_name.name} is not cached, fetching #{resource_type}"
         gem_fetcher.fetch(gem_name.id, resource_type) do |content, properties|
-          gem = gem_resource.save({ resource_type => content }, headers: { resource_type => properties })
+          resource_properties = {
+            upstream: upstream.to_s,
+            gem_name: gem_name.name,
+            headers: { resource_type => properties }
+          }
+
+          gem = gem_resource.save({ resource_type => content }, resource_properties)
           Gemstash::DB::CachedRubygem.store(upstream, gem_name, resource_type)
           gem
         end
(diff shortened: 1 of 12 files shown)


lib/gemstash/gem_source/upstream_source.rb
```diff
@@ -160,7 +160,15 @@ def fetch_local_gem(gem_name, gem_resource, resource_type)
...
         gem_fetcher.fetch(gem_name.id, resource_type) do |content, properties|
-          gem_resource.save({ resource_type => content }, headers: { resource_type => properties })
+          resource_properties = {
+            upstream: upstream.to_s,
+            gem_name: gem_name.name,
+            headers: { resource_type => properties }
+          }
+
+          gem = gem_resource.save({ resource_type => content }, resource_properties)
+          Gemstash::DB::CachedRubygem.store(upstream, gem_name, resource_type)
```
pcarranza: Just thinking, could we use an instance instead of a class method here? (just wondering, maybe this is the way of doing things)

lib/gemstash/migrations/03_cached_gems.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    create_table :cached_rubygems do
+      primary_key :id
+      Integer :upstream_id, null: false
+      String :name, size: 191, null: false
+      String :resource_type, size: 191, null: false
+      DateTime :created_at, null: false
+      DateTime :updated_at, null: false
+      index [:upstream_id, :name, :resource_type], unique: true
```
pcarranza: I'm thinking that the resource type will change less than the name, could this index be upstream_id, resource_type, name ?

All the variance will be in the name, right? The resource type will be pretty much constant, or we are either going to get the content or something else, not both at the same time.

lib/gemstash/gem_source/upstream_source.rb
smellsblue: That's typically how I deal with these sorts of methods. I like it that way because I can hide any transactions or nested creations or lookups in the class method, along with extracting the proper information out of the various objects passed in. I think of it like a factory method basically, which seems appropriate as a class method. But to be honest, I'm not sure what would happen if we tried to write a custom initialize method for something that extends `Sequel::Model`...

pcarranza: I'm ok with it. It's just that class (or static) methods usually resist refactoring basically because they are global.

Not a blocker, just a thought.

pcarranza:
Looks good, go for it :shipit:

PR #71, opened 2016-01-23
Added support for mysql2 adapter

chriseckhardt:
This pull request also required an update to .rubocop-bundler.yml, but I left the satisfy-rubocop updates
for another PR to keep this one smaller.


pcarranza:
Looks like rubocop is not happy with the changes, it just refuses to take conditionals at all.

I would suggest to drop the rubocop yml changes to get the CI green, then we can just merge it in.

PR #74, opened 2016-02-19
Allow log_file to be configured, default to server.log in base dir

mrchucho:
We deploy gemstash to our continuous delivery server and would like to have it log to something other than the default location. This small change adds support for a config option ```:log_file``` that overrides the default log file location.

### DIFF
--- a/lib/gemstash/cli/authorize.rb
+++ b/lib/gemstash/cli/authorize.rb
@@ -20,7 +20,7 @@ module Gemstash
     private
 
       def setup_logging
-        Gemstash::Logging.setup_logger(gemstash_env.base_file("server.log"))
+        Gemstash::Logging.setup_logger(gemstash_env.log_file)
       end
 
       def remove_authorization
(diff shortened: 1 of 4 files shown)


mrchucho:
Hrm, seeing the same errors in Travis that I saw when trying to test locally. The errors don't appear to be related to my change, however.

smellsblue:
@mrchucho strangely enough, it might be the new specs... sometimes they leak into each other in weird ways... like a new database connection that causes failures because the DB is wrapped in a transaction. I've had to hack around this sort of thing before. I'll see if I can help out a little bit later today :-)

smellsblue:
@mrchucho nope, that's not it, I see the specs got added after the code, and it failed then too... I'll look closer later...

smellsblue:
OK, so I was running specs and everything was working fine... then I bundle updated and things started breaking... I suspect there is a gem that got updated and is breaking things...

pcarranza:
Hey @mrchucho, I think that the specs got fixed in master, could you rebase this on top of master again please?

After that I think we are good to merge.

mrchucho:
Rebased! Thanks!

PR #77, opened 2016-02-23
Don't pass along the Content-Length header

smellsblue:
With the latest version of `Puma`, [this line](https://github.com/puma/puma/blob/642ca4f6cb9a0cf7e9174809fb0d78357f9e2ee4/lib/puma/server.rb#L613) is giving us problems from [this commit](https://github.com/puma/puma/commit/642ca4f6cb9a0cf7e9174809fb0d78357f9e2ee4).  Specifically, we are setting a `Content-Length` with a downcased header name. Before that commit in `Puma`, the normal cased version of the `Content-Length` header would be the only thing that triggered the line, but now the downcased version will as well. This in turn causes `Puma` to blow up. This is easily fixed by not passing along the `Content-Length` header we stashed from RubyGems.org. I suspect the content length is somehow differing from what should be sent... maybe a different bug in our code, or a bug somewhere in `Puma` when the content length is being sent. Regardless, this fixes the bug appearing in the specs.

Additionally, this was causing a bug for @chriseckhardt where gems would download and appear to be corrupted. This fix in turn fixes that bug as well as the spec failures with the latest `Puma`.

### DIFF
--- a/gemstash.gemspec
+++ b/gemstash.gemspec
@@ -45,5 +45,5 @@ you push your own private gems as well."
   spec.add_development_dependency "rack-test", "~> 0.6"
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.3"
-  spec.add_development_dependency "rubocop", "~> 0.34"
+  spec.add_development_dependency "rubocop", "0.35.1"
 end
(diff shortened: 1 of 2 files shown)


smellsblue:
I think I tracked down why this is problematic.

We are using `Rack::Deflater` in both the [test rack config](https://github.com/bundler/gemstash/blob/master/spec/support/test_gemstash_server.ru#L2) and the gem provided [rack config](https://github.com/bundler/gemstash/blob/master/lib/gemstash/config.ru#L4). This in turn will alter the length of the served content, so the content length header we set from the cached headers will absolutely be wrong. I'm actually a bit surprised this hasn't caused any issues before now.

pcarranza:
:+1:

Issue #79, opened 2016-03-11
Faraday::ConnectionFailed


midwire:
Howdy, and thanks for this great gem caching service!

Part of the reason I decided to try gemstash is because I'm on a slow T1 connection. So when I try to pull some larger gems I sometimes get:

`ERROR - Connection failure - execution expired (Faraday::ConnectionFailed)`

It would be great to be able to set an increased timeout value for Faraday.  If I get a few mins. I'll drop in a PR here.

Thanks again!

smellsblue:
Thanks so much for trying it out and the feedback! I'd be happy to merge it in when you have a PR ready! Would it help to get a patch release of the gem, or are you running Gemstash directly from a git clone?

midwire:
I'm running from a git clone at the moment. Just working on the config setup portion now, as I've already got the patch itself working and passed all specs. Calling the option `config[:fetch_timeout]`, but please let me know if you want to change it, or feel free to.

pcarranza:
I wonder,

If we catch the error as a timeout and not a general connection error, does it make sense to have some form of exponential retry? that would adjust itself when we have a gem that is too big and we are in a slooooow connection.

Something like leaving the default timeout configured, no idea how much is it, say 5 seconds; catch the timeout error, double the timeout to 10 seconds and retry. If it fails again this second time we could just give up and let the error bubble up or double and try again one more time.

That would be a bit more dynamic and will allow us to not think so much about the connection we are dealing with. And I love not having to think at all.

Just a thought, the PR is a good solution also.

midwire:
I usually prefer exponential backoff as the solution for things like this. The issue that I foresee is what if the 3 backoffs isn't enough and it still times out? Do we do 4? And then what if someone is on dial-up... LOL. I jest but, also like the ability to just set it and forget it. I could be on the beach somewhere with horrible bandwidth, set it to 999, and have some drinks while I wait for it to cache up 'nokogiri' and all the other gems for my app, and not have to worry about it timing out. :watch: 

Re: not having to think about it, we could make it a hidden config option and not ask the question during 'setup'?

I'm willing to try exponential if you'd rather have that. It may take me some time to get to it as I've got to wrap up for tonight. Let me know, and I'll pursue it this coming week.

pcarranza:
Oh no, I think that having the ability to set it up as a general timeout is great. Do not ditch that.

[...]

PR #81, opened 2016-03-12
Make gem fetch timeout configurable

midwire:
* Referencing: #79 (issue)
* Ask for timeout value during setup
* Add default timeout value of 20


spec/gemstash/http_client_spec.rb
```diff
@@ -20,6 +20,16 @@
...
   end
 
+  describe "#for" do
+    it "sets configured timeout value" do
+      configured_timeout = 99
+      client = Gemstash::HTTPClient.for(Gemstash::Upstream.new(@server.url), configured_timeout)
+      expect(
+        client.instance_variable_get(:@client).options.timeout
+      ).to eq(configured_timeout)
+    end
```
pcarranza: Could this be tested as a behaviour rather than as cracking the object open by reflection to check that the internal have a specific shape?

What I mean is that in this file, at the beginning we are setting up a SimpleServer, we could have a "slow mount" that takes something like 0.2 seconds sleep and then setup the upstream with a timeout of 0.1, that should be enough to make it timeout. Then a second test with a timeout of 0.3 should be enough.

Consider that I am making times up and that it could be a bit harder to fine tune the testing, but I would be much more happier checking behaviour than implementation.

lib/gemstash/http_client.rb
```diff
@@ -26,10 +26,11 @@ class HTTPClient
 
     DEFAULT_USER_AGENT = "Gemstash/#{Gemstash::VERSION}"
 
-    def self.for(upstream)
+    def self.for(upstream, timeout = 20)
       client = Faraday.new(upstream.to_s) do |config|
         config.use FaradayMiddleware::FollowRedirects
         config.adapter :net_http
+        config.options.timeout = timeout
       end
```
pcarranza: Call it fetch_timeout for symmetry with the configuration variable and maybe consider making it a keyword argument?

spec/spec_helper.rb
```diff
@@ -64,6 +64,10 @@
     TestGemstashServer.join_all
   end
 
+  # Tag examples with focus: true to run only those
+  config.filter_run :focus
+  config.run_all_when_everything_filtered = true
+
```
pcarranza: Forgotten?

pcarranza:
Thanks so much for the PR :)

I left a few thoughts for your consideration.

lib/gemstash/http_client.rb
midwire: Sure. 'keyword' as in an options hash?

spec/gemstash/http_client_spec.rb
midwire: I totally agree re: testing behavior. My laziness is showing. I'll give it a shot.

spec/spec_helper.rb
midwire: Yep, I'll remove that.

spec/gemstash/http_client_spec.rb
pcarranza: I feel you :laughing:

lib/gemstash/http_client.rb
pcarranza: `fetch_timeout: 20`kinda, I think that is the options hash, yes.

bianjp:
Any progress?

midwire:
sorry, i've been pounded with work and forgotten actually. I'll try to get to this tonight.

spec/gemstash/http_client_spec.rb
midwire: @pcarranza FYI: It is going to be quite a hack to force the `SimpleServer` to sleep during the `get`. The problem with a "slow mount" method is that it will pause or `sleep` during the mount, not during the get. So it will merrily sleep for x number of seconds before defining the path you want to mount, then we finally get to the actual `get` call where we expect Faraday (NET::HTTP actually) to timeout. The sleep already happened further up the stack in the `slow_mount` call.

So in order to force a sleep during the `get` request, we need to break open `Servlet. The `Servlet` class gets passed to `WEBrick::HTTPServer.mount` which calls `initialize` on `Servlet` so we can't directly pass a timeout value to the servlet. We could define a `SlowServlet` class but then we have to parameterize it at the `SimpleServer` level by either making a `SlowSimpleServer` as well or passing an argument to `initialize`.

Hopefully all this makes sense. In other words it is a major PITA to try and setup the test to actually make a real `get` call and timeout using the passed fetch_timeout value. Not sure how you want to proceed. Let me know.

midwire: PS: I pushed a sample commit that does what I describe above. It isn't as elegant as I'd prefer but it tests behavior.

pcarranza:
LGTM, @smellsblue any objection?

lib/gemstash/http_client.rb
```diff
@@ -26,10 +26,11 @@ class HTTPClient
 
     DEFAULT_USER_AGENT = "Gemstash/#{Gemstash::VERSION}".freeze
 
-    def self.for(upstream)
+    def self.for(upstream, timeout = 20)
       client = Faraday.new(upstream.to_s) do |config|
         config.use FaradayMiddleware::FollowRedirects
         config.adapter :net_http
+        config.options.timeout = timeout
```
smellsblue: How about fetch the config timeout here instead of passing it in? So `config.options.timeout = Gemstash::Env.current.config[:fetch_timeout]`. You could also use the helper, so in `Gemstash::HTTPClient` it could `extend Gemstash::Env::Helper` and then `Gemstash::HTTPClient.for` could do `config.options.timeout = gemstash_env.config[:fetch_timeout]`

lib/gemstash/web.rb
```diff
@@ -18,7 +18,8 @@ def initialize(gemstash_env: nil, http_client_builder: nil)
     end
 
     def http_client_for(server_url)
-      @http_client_builder.for(server_url)
+      timeout = Gemstash::Configuration.new[:fetch_timeout] || 20
+      @http_client_builder.for(server_url, timeout)
```
smellsblue: This won't necessarily get the right config, you need to use the config in the environment, so it would be `@gemstash_env.config[:fetch_timeout]` or `Gemstash::Env.current.config[:fetch_timeout]`. Is there a need to vary the timeout from the config? If not, I think it might be nicer if the `Gemstash::HTTPClient` fetched the timeout from the config directly :-)

There should be no need to hard code the default, the config object handles that via `Gemstash::Configuration::DEFAULTS`.

midwire:
Cool. Thanks for the help @smellsblue. Changes are applied.

smellsblue:
:+1: Thanks for all the work!

PR #86, opened 2016-03-31
Support ERB parsing in YAML config

jiexinhuang:
Ideally, we don't want to have configuration hardcoded and checked in source control.
This change will allow users to fetch config from environment variables and make it easier to deploy Gemstash into different environments.

### DIFF
--- a/lib/gemstash/configuration.rb
+++ b/lib/gemstash/configuration.rb
@@ -1,4 +1,5 @@
 require "yaml"
+require "erb"
 
 module Gemstash
   #:nodoc:
@@ -31,7 +32,7 @@ module Gemstash
       file ||= DEFAULT_FILE
 
       if File.exist?(file)
-        @config = YAML.load_file(file)
+        @config = YAML.load(::ERB.new(File.read(file)).result)
         @config = DEFAULTS.merge(@config)
         @config.freeze
       else


smellsblue:
This looks pretty neat! Could you add some documentation to (docs/config.md)[https://github.com/bundler/gemstash/blob/master/docs/config.md] and create `spec/gemstash/configuration_spec.rb` with a test or 2?

Would there be any potential downfall to always running the config YAML file through ERB (security, confusion, or conflict with YAML syntax)? Does it at all make sense to make it more comlex, like only run through ERB if `config.yml` doesn't exist but `config.yml.erb` does? I know there's precedence with database.yml for Rails config, so maybe that's unnecessary?

indirect:
Personally, I like the idea of being able to opt in to erb processing with `config.yml.erb`! That seems like a clean way to make sure that everyone knows what's going on.

rjocoleman:
I'm keen to help get this feature in. What's needed to land this?

* Relevant tests
* Erb parsing of `config.yml.erb` file only
* What if `config.yml` and `config.yml.erb` both exist?

rjocoleman:
@smellsblue @indirect I've made some changes into a new PR #90 based on the feedback here.

smellsblue:
Closing this as it has been merged in via #90

PR #87, opened 2016-04-16
Add better error messaging when the resources do not have a version recorded

pcarranza:
I just run into this, my cache was too old and gemstash crashed with a _Nil does not define the method <=_

So I just added a simple way of logging a meaningful message.

cc/ @smellsblue

### DIFF
--- a/lib/gemstash/storage.rb
+++ b/lib/gemstash/storage.rb
@@ -275,7 +275,9 @@ module Gemstash
     end
 
     def check_resource_version
-      version = @properties[:gemstash_resource_version]
+      version = @properties.fetch(:gemstash_resource_version) do
+        raise "Resources are too old and don't have version, please wipe the cache directory"
+      end
       return if version <= Gemstash::Resource::VERSION
       reset
       raise Gemstash::Resource::VersionTooNew.new(name, folder, version)


smellsblue:
This error will only happen if you ran Gemstash before the first release, so I think we are the only ones likely to benefit from this, unless a bug causes the version to not be included.

With that in mind, maybe the error message should indicate that there is something wrong with the stored resource metadata?

It seems like it might also be useful to indicate the directory the user should delete to fix the issue, however deleting all the cache seems drastic given this is most likely to occur due to a bug. Perhaps point to the specific gem causing the problem and request it be deleted?

olleolleolle:
The first bit: 

`"Resources are too old and don't have version, please wipe the cache directory"`

into something like -> 

`"Stored metadata for resource #{name.inspect} lacks gemstash_resource_version. Please delete that gem."`

(Still needs the directory reference in there.)

hmistry:
@olleolleolle That sounds very technical where I, as a user, wouldn't understand what it means if I'm not super intimate with gemstash. I personally prefer the current simpler actionable message though I think it can be improved by clarifying what resources are (are they gems, metadata, both?) and which particular resource is affected. The latter may not matter if you're doing a blanket wipe of the cache directory. I find error messages are best when they are specific, concisely explain the issue in simple terms, and actionable e.g. "The installed rack gem is too old. Please delete the gem located at the following directory: /path/gem and try running the command again."

olleolleolle:
How about:

```ruby
      version = @properties.fetch(:gemstash_resource_version) do
        raise "The installed #{name.inspect} gem has outdated metadata. " \
               "Please delete the gem located at the following directory: " \
               "#{folder} and try running the command again."
      end
```

olleolleolle:
Or, inspired by the lingo in the other, more specific Version error:

```ruby
      version = @properties.fetch(:gemstash_resource_version) do
        raise "Gemstash resource version #{Gemstash::Resource::VERSION} does " \
              "not support the outdated metadata for resource " \
              "#{name.inspect} found at #{folder}."
      end
```

olleolleolle:
@pcarranza Hi! Are you interested in changing the text of the error message?

If not, would you be OK with me adding another PR which does that change that @hmistry thumbs-up'd?

PR #88, opened 2016-05-10
Add fallback timeout documentation for when Gemstash's off/missing

parndt:
Per bundler/bundler#4186 this documents how to add a fallback timeout to prevent this happening when Gemstash is off:

```
$ bundle update
Fetching source index from http://localhost:9292/
Retrying fetcher due to error (2/4): Bundler::HTTPError Could not fetch specs from http://localhost:9292/
Retrying fetcher due to error (3/4): Bundler::HTTPError Could not fetch specs from http://localhost:9292/
Retrying fetcher due to error (4/4): Bundler::HTTPError Could not fetch specs from http://localhost:9292/
Could not fetch specs from http://localhost:9292/
```

After configuring a fallback, as described:

```
$ bundle config mirror.https://rubygems.org.fallback_timeout true
```

Instead, after trying Gemstash first and quickly failing, we see this:

```
$ bundle update
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Fetching dependency metadata from https://rubygems.org/
```

So we can see that it skips to using https://rubygems.org instead. 

Now, Gemstash seems much more like a viable option for a mirror!

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,24 @@ $ rm -rf Gemfile.lock .bundle
 $ bundle
 ```
 
+### Falling back to rubygems.org
+
+If you want to make sure that your bundling from https://rubygems.org still
+works as expected when the Gemstash server is not running, you can easily 
+configure Bundler to fallback to https://rubygems.org.
+
+```
+bundle config mirror.https://rubygems.org.fallback_timeout true
+```
+
+You can also configure this fallback as a number of seconds in case the Gemstash
+server is simply unresponsive.
+This example uses a 3 second timeout:
+
+```
+bundle config mirror.https://rubygems.org.fallback_timeout 3
+```
+
 ### Stopping the Server
 
 Once you've finish using your Gemstash server, you can stop it just as easily as


pcarranza:
Glad you like it, I coded that feature myself 😄 

We were missing documentation for this 👍

parndt:
I suppose I should have set `[ci skip]`..

@pcarranza would you also like documentation [for bundler.io](https://github.com/bundler/bundler-site/blob/784e03ff9d46bafc8608a10e5dacab82f1463918/source/v1.12/bundle_config.haml#L264)?

pcarranza:
@parndt Heh, I thought I did added documentation, but I did not.

Please do, it will be great to have it properly documented.

Don't worry about CI, this is just documentation and JRuby complaining is not something odd, sadly.

pcarranza:
I'm merging with CI failing because it is not a code change, but a documentation change.

parndt:
@pcarranza @indirect @smellsblue I've also documented this for Bundler's documentation site over at bundler/bundler-site#212

indirect:
Thanks @parndt!

smellsblue:
:+1:

PR #92, opened 2016-05-21
Improve code climate

smellsblue:
Add a code climate badge to the README and fix (hopefully) all of the issues they found

PR #93, opened 2016-05-21
Refactor authorization

smellsblue:
Refactor authorization. This includes #92 along with a refactoring to isolate authorization in a way to be easily replaceable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![Build Status](https://travis-ci.org/bundler/gemstash.svg?branch=master)](https://travis-ci.org/bundler/gemstash)
+[![Code Climate](https://codeclimate.com/github/bundler/gemstash/badges/gpa.svg)](https://codeclimate.com/github/bundler/gemstash)
 
 <p align="center"><img src="gemstash.png" /></p>
 
@@ -82,7 +83,7 @@ $ bundle
 ### Falling back to rubygems.org
 
 If you want to make sure that your bundling from https://rubygems.org still
-works as expected when the Gemstash server is not running, you can easily 
+works as expected when the Gemstash server is not running, you can easily
 configure Bundler to fallback to https://rubygems.org.
 
 ```
--- a/docs/config.md
+++ b/docs/config.md
@@ -116,6 +116,19 @@ This maps directly to the [Puma bind
 flag](https://github.com/puma/puma#binding-tcp--sockets), and will support
 anything valid for that flag.
 
+## Protected Fetch
+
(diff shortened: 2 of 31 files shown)


rjocoleman:
@smellsblue would named parameters make sense for the `serve` methods? In the context of something like `Gemstash::SpecBuilder` I was thinking of creating a boolean param for `prerelease` then using the generic serve method.

`Gemstash::SpecsBuilder.serve(nil, {}, {prerelease: true})` vs `Gemstash::SpecsBuilder.serve(config: {prerelease: true})` or `Gemstash::SpecsBuilder.serve(auth_key: "foo")`

smellsblue:
@rjocoleman well, I kind of think such options should be wrapped in the object before it gets to the `serve` method. So in the `Gemstash::SpecsBuilder` case, I think it might be better if the object passed to `authenticated` were an instance rather than a class (so you could pass `Gemstash::SpecsBuilder.new(prerelease: true)`), and then the `serve(auth, request, params)` method would be an instance method rather than a class method.

My intent for the `serve` method is that it has all it needs.... but as I think this through more, I think I can drop all the arguments in favor of just the Sinatra app object (well... private gem source that delegates to a Sinatra app).

The ultimate goal of this refactoring was to pave the way for the authorization setup to be replaceable with other authentication/authorization via a plugin, specifically LDAP (which someone had requested in the slack channel). I'm hoping it just has raw access to everything so the plugin could do much more complicated things if necessary.

Anyways, I think I have another commit to do for this now....... thanks for getting me to think a little deeper on this :-)

rjocoleman:
This is half baked but what about something like this: https://github.com/bundler/gemstash/compare/refactor-authorization...rjocoleman:refactor-authorization?expand=1

rjocoleman:
Let's get a `.codeclimate.yml` file in here too so it can be used with the [CLI](https://github.com/codeclimate/codeclimate) locally.

Potentially: 

```yaml
---
engines:
  bundler-audit:
    enabled: true
  duplication:
    enabled: true
    config:
      languages:
      - ruby
  fixme:
    enabled: true
  rubocop:
    enabled: true
ratings:
  paths:
  - Gemfile.lock
  - "**.rb"
exclude_paths:
- spec/
```

rjocoleman:
@smellsblue Are there any further blockers here you need assistance with?

smellsblue:
@rjocoleman I think I'm good with this as is, but I was hoping to get @pcarranza to review it before I merged it in. If you would like to do the reviewing on what I've done, I think I'd be ok merging it in now :-)

pcarranza:
Sorry, been extremely busy. Will review today.

PR #94, opened 2016-05-21
Expanded - Refactor authorization - Add protected fetch for private gems

### LINKED ISSUE #24: Protected fetch
We have push behind an API key, as will be yank and unyank.

Should fetching private gems support the key as well?

Will bundler send API key from rubygems config?

Should we add that to bundler if not?

Maybe we send a 401 and bundler can look for API key from rubygems config? Maybe specify API key name in config and it will use it automatically?

rjocoleman:
This builds on #92, #93 and resolves #24 

@smellsblue I think this covers off all your feedback - thanks!

This PR contains:

* Iteration on @smellsblue's work on the Authorization refactor
* `protected_fetch` (default false) in `Gemstash::Configuration`
* Configuration of `protected_fetch` via `$ gemstash setup`
* A new `fetch` permission for Authorization
* Authentication via HTTP Basic Auth (token as username) to support `gem`, `bundler` and uri based auth in `Gemfiles`
* Docs to suit.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -116,6 +116,19 @@ This maps directly to the [Puma bind
 flag](https://github.com/puma/puma#binding-tcp--sockets), and will support
 anything valid for that flag.
 
+## Protected Fetch
+
+Gemstash by default allows un-authenticated access for Private gems. Authenticated access is available via the `:protected_fetch` configuration key.
+
+
+```yaml
+# ~/.gemstash/config.yml
+---
+:protected_fetch: true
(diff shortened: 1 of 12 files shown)


rjocoleman:
@smellsblue if this is acceptable I'll do the `serve_gem` and `serve_marshal` methods in a similar way and stick them in this PR too.

rjocoleman:
@smellsblue Pending any further feedback, this should be good to go

PR #102, opened 2016-07-07
Redirect /versions and /info/* to index.rubygems.org

### LINKED ISSUE #100: /versions and /info/* are redirected to rubygems.org
Gemstash is currently redirecting requests for `/versions` to https://rubygems.org/versions instead of the correct https://index.rubygems.org/versions. The endpoint for `/info/*` should also be updated to redirect to https://index.rubygems.org/info/*.

```
bundle update --verbose
HTTP GET https://internal.gemstash.server/versions
HTTP 302 Found
HTTP GET https://rubygems.org/versions
HTTP 200 OK
HTTP GET https://internal.gemstash.server/versions
HTTP 302 Found
HTTP GET https://rubygems.org/versions
HTTP 200 OK
The checksum of /versions does not match the checksum provided by the server! Something is wrong (local checksum is "\"f3e85d51952e0136d9a61006a82ab686\"", was expecting "W/\"f3e85d51952e0136d9a61006a82ab686\"").
HTTP GET https://internal.gemstash.server/api/v1/dependencies
HTTP 200 OK
```

This also has the side effect of causing a checksum mismatch due to rubygems.org serving weak ETags for the `/versions` endpoint. See https://github.com/bundler/bundler/issues/4764 for reference.

smellsblue:
This fixes #100 
This also includes #101 since it should be included for consistency and correctness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@
 
 ## What is Gemstash?
 
-Gemstash is both a cache for remote servers such as https://www.rubygems.org,
-and a private gem source.
+Gemstash is both a cache for remote servers such as https://rubygems.org, and a
+private gem source.
 
 If you are using [bundler](http://bundler.io/) across many machines that have
 access to a server within your control, you might want to use Gemstash.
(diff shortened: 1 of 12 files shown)


indirect:
@smellsblue this looks great! 👍

PR #109, opened 2016-10-13
Better documentation

### LINKED ISSUE #96: Documentation and release sync
As mentioned on Slack..

Docs in master do not represent the currently available version.
At very least we can add a note to this effect and a link to releases?

As new features are added and documented a reference to their targeted version too?
eg "ERB support in YAML configuration file (0.2 and greater)" (although this means contributors need to have an idea of version numbers/releases or it's more work on release)

smellsblue:
This PR makes the following changes:
* Refactor docs a bit, in particular, break out the reference into separate documents per command
* Embed docs in the gem, and use `man` to output the docs when `gemstash help` is run, and embedded txt files when `man` is unavailable
* Use `pandoc` to generate all docs from a single source (it generates the new embedded docs, the github markdown docs, and the soon to come html docs)
* Drop the table of contents Rake task, since the new docs structure doesn't need it anymore
* Auto generate the docs when building the gem
* Reduce contents of the built gem to just the `lib`, and `exe` dirs, the license, changlog, and code of conduct

This fixes #96


indirect:
Prettttttty great. 👍  from me!

PR #110, opened 2016-10-14
Fix JRuby failures

smellsblue:
* Fix the JRuby test failures
* Fix integration tests so they can run in isolation
* Commit changes Bundler made to test config files


indirect:
Looks good.

PR #121, opened 2016-11-12
Travis: add Gemfile for ruby-head

olleolleolle:
This PR tries to make Ruby 2.4.0 build right.

- add a `gemfiles/` folder with a Ruby 2.4.x-specific Gemfile that uses `json` gem in v2.x
- point to it in `.travis.yml`
- git ignore lockfiles in there

### DIFF
--- a/gemstash.gemspec
+++ b/gemstash.gemspec
@@ -28,7 +28,7 @@ you push your own private gems as well."
   spec.executables   = spec.files.grep(%r{^exe/}) {|f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  if RUBY_VERSION > "2.3"
+  if RUBY_VERSION > "2.4"
     spec.add_runtime_dependency "activesupport", "~> 5.0"
   else
     spec.add_runtime_dependency "activesupport", "~> 4.2"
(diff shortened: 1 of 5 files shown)


gemstash.gemspec
```diff
@@ -28,7 +28,12 @@ you push your own private gems as well."
   spec.executables   = spec.files.grep(%r{^exe/}) {|f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "activesupport", "~> 4.2"
+  if RUBY_VERSION > "2.3"
```
smellsblue: How is this going to affect building and publishing the gem? Does this mean the runtime dependencies will be based on which version of Ruby I am running when I build the gem?

olleolleolle: Yes, sadly.

Perhaps we could make the check match 2.4-only, since that's the one which fails hard on older activesupport.

Like so:

```ruby
if RUBY_VERSION > "2.4"
```

smellsblue: How about if we make the dependency:
```
spec.add_runtime_dependency "activesupport", ">= 4.2"
```
That way it should work on both older and newer versions of Ruby?

olleolleolle: We can try it out and hope!

Those who want to keep `activesupport` back on their installations can use a Gemfile should they need to.

PR #131, opened 2016-12-26
Support the latest_specs endpoint

### LINKED ISSUE #129: Several commands fail for private gems
I just configured gemstash as a private gem server for a project.  Our intention is to use this in place of geminabox.

During my initial testing, everything went fine.  I was able to import all of the versions of our private gems, and I've been able to push new gems just fine.  Running a `bundle install` or `bundle update` works as well.

The first indication of a problem is that `gem install my_private_gem` fails with:

```
ERROR:  Could not find a valid gem 'my_private_gem' (>= 0) in any repository
```

Digging further, I found that `gem list --clear-sources -r -s http://my-gem-server:9292/private` returned no results, but `gem list --clear-sources -a -r -s http://my-gem-server:9292/private` lists all of the versions of all of our private gems as expected.

I also found similar results with the `gem search` command: without the `-a` flag, no results are returned; with it, everything works as expected.

I looked in my server logs, and I found that the successful commands are hitting `/private/specs.4.8.gz`, but the unsuccessful commands are hitting `/private/latest_specs.4.8.gz` instead.

Looking in the code, it appears that there is not yet an implementation for the `latest_specs.4.8.gz` endpoint in `lib/gemstash/gem_source/private_source.rb`.

Do I understand the situation correctly?  If so, it seems like this is a fundamental limitation of using gemstash as a private gem server in its current state, and that the limitation should be clearly documented.

Is there a workaround for this issue?  If not, we might have to back away from using gemstash as our private gemserver.  I'd hate to do that, because I really like gemstash so far.

What would be involved with supporting the latest_gems endpoint?  How hard would it be for someone unfamiliar with the code (me) to implement this?  If there's too much context required, is there anything I can do to help?

randycoulman:
This adds support for the `latest_specs.4.8.gz` endpoint that is required for several gem commands to work properly (e.g. `gem install` and `gem search` without the `-a` flag).  See #129 for more details.

There might be a more efficient way to implement the database query.  I'm not aware of a way to do proper gem version comparisons in the DB, so it seemed to just pull all of the versions and do the latest version selection in Ruby.

If a gem has platform-specific versions, the latest version for each platform will be returned by this code.  That seems to match what the `latest_specs.4.8.gz` endpoint on rubygems.org does.

I added several unit tests and one new integration test that shows the difference in behavior between `gem search -a` and `gem search`.

It might be worth adding integration tests for the `gem install` command, but I wasn't sure what kind of test infrastructure that would require.  I'd be happy to write such a spec if someone can give me a little guidance there.

Fixes #129

### DIFF
--- a/lib/gemstash/db/version.rb
+++ b/lib/gemstash/db/version.rb
@@ -30,8 +30,17 @@ module Gemstash
         end
       end
 
-      def self.for_spec_collection(prerelease: false)
-        where(indexed: true, prerelease: prerelease).association_join(:rubygem)
+      def self.for_spec_collection(prerelease: false, latest: false)
+        versions = where(indexed: true, prerelease: prerelease).association_join(:rubygem)
+        latest ? select_latest(versions) : versions
+      end
+
+      def self.select_latest(versions)
+        versions.
+          all.
+          group_by {|version| [version.rubygem_id, version.platform] }.
+          values.
+          map {|gem_versions| gem_versions.max_by {|version| Gem::Version.new(version.number) } }
       end
 
       def self.find_by_spec(gem_id, spec)
--- a/lib/gemstash/gem_source/private_source.rb
+++ b/lib/gemstash/gem_source/private_source.rb
@@ -80,12 +80,13 @@ module Gemstash
         end
       end
 
-      def serve_latest_specs
-        halt 403, "Not yet supported"
-      end
-
       def serve_specs
         params[:prerelease] = false
+        protected(Gemstash::SpecsBuilder)
+      end
+
+      def serve_latest_specs
+        params[:latest] = true
         protected(Gemstash::SpecsBuilder)
       end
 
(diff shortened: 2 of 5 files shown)


indirect:
Awesome, thanks for writing this!

As you noticed, only `specs.4.8.gz` is actively used by Bundler, and only when the `--full-index` option is passed. Today, Bundler tries to use the compact index provided at `/versions` and `/info/GEMNAME`, then falls back on the dependency API available at `/api/v1/dependencies` and only if that is not available will it fall back on `specs.4.8.gz`.

I believe these three files are the only ones currently used by RubyGems: https://github.com/rubygems/rubygems.org/blob/cb09831cd8e827a821b8cdc203fecd8bbab6b722/app/middleware/hostess.rb#L6-L8.

indirect:
Sorry for the confusion! It is needed, because rubygems uses it. :)

smellsblue:
@randycoulman Thanks so much for the work! Sorry for being MIA, but I'm going to try to backport this to the `1.0` branch and get it released ASAP. I'll also try to add the integration test you mentioned.

randycoulman:
@smellsblue Awesome, thanks!  Let me know if I can be any help with the backport or the integration test.

PR #147, opened 2017-05-18
Travis: jruby-9.1.12.0

olleolleolle:
This PR updates the CI matrix to use latest JRuby.

http://jruby.org/2017/06/15/jruby-9-1-12-0.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
     - rvm: 2.4.1
     - rvm: ruby-head
       gemfile: 'gemfiles/Gemfile.ruby-head'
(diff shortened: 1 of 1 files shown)


olleolleolle:
@smellsblue Can you review this?

jules2689:
LGTM. @indirect should we declare a "supported versions" document like [we did for Bundler](https://github.com/bundler/bundler/issues/5789)?

indirect:
@jules2689 we probably should, yeah. I typically want to aim for "the currently supported Rubies", dropping active support at the same time as the Ruby goes into "security patches only".

@bundlerbot r+

PR #160, opened 2017-08-16
Allow pidfile option

sinneduy:
so you can do `--pidfile <some pidfile here>`


sinneduy:
accidental pr, will open a new one when its ready

olleolleolle:
Use RUPYOPT=-rpry to start Ruby with pry already required! It's a way to not have to add it to gemspec etc.

PR #179, opened 2018-04-29
Update sinatra dependency to >= 1.4, < 3.0

bronzdoc:


bronzdoc:
@bundlerbot r+

greysteil:
👍. I *think* Dependabot would run on this repo fine, if you want to automate these.

We might want to drop support for Ruby 2.0 and Ruby 2.1 anyway - is there a good reason for supporting them on this gem? They're [end of life](https://www.ruby-lang.org/en/downloads/branches/).

bronzdoc:
@greysteil I'n not %100 if there's a good reason, @indirect what you think?

PR #172, opened 2017-09-12
gem: thor v0.20.0

olleolleolle:
This PR updates `thor` to the next minor version.

[Changelog](https://github.com/erikhuda/thor/blob/master/CHANGELOG.md#0200)

PR #180, opened 2018-04-30
gemspec: use Thor 0.20.0

olleolleolle:
#172 got confused.

This is the same change: use latest available Thor.


bronzdoc:
Thanks @olleolleolle !

@bundlerbot r+

PR #183, opened 2018-06-04
do not set puma worker config when running jruby

andreaseger:
puma doesn't like if the worker config is set on jruby and won't even start

### DIFF
--- a/lib/gemstash/puma.rb
+++ b/lib/gemstash/puma.rb
@@ -2,5 +2,5 @@ require "gemstash"
 
 threads 0, Gemstash::Env.current.config[:puma_threads].to_i
 bind Gemstash::Env.current.config[:bind].to_s
-workers Gemstash::Env.current.config[:puma_workers].to_i unless RUBY_PLATFORM == 'java'
+workers Gemstash::Env.current.config[:puma_workers].to_i unless RUBY_PLATFORM == "java"
 rackup Gemstash::Env.current.rackup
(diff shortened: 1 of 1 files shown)


olleolleolle:
Do we need to teach this difference to folks in the configuration documentation?

andreaseger:
AFAICS the docs arn't mentioning puma worker at all right now. In the lastest released version (1.1.0) this also didn't really matter as it was always set to `1`. In master I now saw that this was made configurable but that was done by someone else.

In general I guess adding a link to the puma config is probably enough - it's explained quite well over there https://github.com/puma/puma#configuration

---

Aside from that any idea why the build fails with `2.3.4` on travis?

olleolleolle:
@andreaseger In order to track this issue, I filed a local bug report for our repository, since perhaps _maybe_ these are issues which occur in our code, when using `config.after(:suite)`?

bronzdoc:
@bundlerbot r+

olleolleolle:
(The failure was "unable to connect to GitHub", a network fluke.)

indirect:
@olleolleolle just sent you an invite to the gemstash team

@bundlerbot retry

notriddle:
That branch is just used by CI under the hood (normally, you'd configure Travis to not build it).

It's regular `staging` that should actually have been run, which is why bundlerbot linked to [those build results](https://travis-ci.org/bundler/gemstash/builds/441546307?utm_source=github_status&utm_medium=notification).

```
  bundle install against gemstash
    with default upstream gems
      behaves like a bundleable project
NameError: uninitialized constant OpenSSL::SSL::SSL2_VERSION
                                                     const_missing at org/jruby/RubyModule.java:3526
```

bronzdoc:
@bundlerbot retry

bronzdoc:
@bundlerbot r+

Issue #185, opened 2018-06-08
Health check may start failing consistently erroneously on database connection death


kurtzur:
Hello,

We've been using this project in production for several months at Gusto. We're using Postgres 9.6.6 as its backend. We recently ran into a problem with the health check (`/health`, not `/health/heartbeat`). All of the instances in our gemstash cluster started failing their health checks with the following message:
```
{"status":{"heartbeat":"OK","storage_read":"OK","storage_write":"OK","db_read":"OK","db_write":"PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint \"rubygems_name_index\"\nDETAIL:  Key (name)=(health_check:fake_gem_name) already exists.\n"}}
```
After reading into this for a while, we determined that one of the gemstash processes (and, in turn, its database connection) died in the middle of running [this code](https://github.com/Gusto/gemstash/blob/master/lib/gemstash/health.rb#L34..L41). Specifically, we think it died after executing `insert`, and before `raise Sequel::Rollback`. The default behavior of Postgres in this case is to commit the transaction. Because the health check code writes a static value (`"health_check:fake_gem_name"`) to a column with a unique index (`name`), we started seeing the failures I specified above. It seems [at least one other person has experienced this issue](https://github.com/bundler/gemstash/pull/146#issuecomment-308032680).

Given the possibility of the health check's DB write being persisted (at least in one database backend), we think `Gemstash::Health.check_db_write` should be made idempotent. Perhaps the following SQL would fit:
```sql
CREATE TABLE IF NOT EXISTS health_check_tests (name varchar(40), count integer);
TRUNCATE health_check_tests;
INSERT INTO health_check_tests VALUES ('test_string', 1);
```

We'd be happy to implement this change. We just wanted to check for feedback first via this issue.

PR #186, opened 2018-06-19
Add health_tests table, and use in Gemstash::Health for idempotency

kurtzur:
Solves https://github.com/bundler/gemstash/issues/185


indirect:
Nice work, everyone 😁

Issue #188, opened 2018-06-21
Gemstash allows overriding published gems


mzruya:
Pushing an existing version of a gem fails the request (see https://github.com/bundler/gemstash/issues/99), This might imply that the operation failed, but it actually succeeds, the underlying gem code is replaced.

Pushing a gem involves:
1. Storing it (https://github.com/bundler/gemstash/blob/master/lib/gemstash/gem_pusher.rb#L31)
2. a DB transaction (https://github.com/bundler/gemstash/blob/master/lib/gemstash/gem_pusher.rb#L72)

The gem storage operation can succeed (overriding the actual code), but the database transaction fails, raising an `ExistingVersionError`.

### Option 1
We could perform the file system operation at the end of the database transaction.

#### Pros
- If the file system operation fails, the database transaction will rollback.
- If the database operations fails, an exception would be raised which would prevent the file system operation from happening

#### Cons
- If the database transaction fails to commit, we'll still override the file system storage.
- This solution requires ensuring that the file system operation is the last and single statement in the database transaction, this would be hard to enforce over time.

### Option 2
Don't override the file, check if it exists before persisting and raise if it does.

### Option 3
Store file names based on a unique identifier.

#### Pros
- This addresses the issue by ensuring files are never overridden by the server.

#### Cons
- Files will be left in storage, we could attempt to cleanup if an exception is raised, but it's not guaranteed.

Options 2 and 3 feel like the most robust approach.
I'm interested in fixing it, any thoughts on the best route?

bronzdoc:
Thanks for bringing this issue at sigh,  i'll ivestigate and try to address it.

PR #190, opened 2018-10-13
Pushing an existing version of a gem should not overwrite the gem con…

bronzdoc:
closes https://github.com/bundler/gemstash/issues/188

### DIFF
--- a/lib/gemstash/gem_pusher.rb
+++ b/lib/gemstash/gem_pusher.rb
@@ -53,6 +53,14 @@ module Gemstash
     end
 
     def store_gem
+      resource_exist = storage.resource(full_name).exist?
+      resource_is_indexed =
+        storage.resource(full_name).properties[:indexed] if resource_exist
+
+      if resource_exist && resource_is_indexed
+        raise ExistingVersionError, "Cannot push to an existing version!"
+      end
+
       storage.resource(full_name).save({ gem: @content }, indexed: true)
     end
 
--- a/lib/gemstash/gem_pusher.rb
+++ b/lib/gemstash/gem_pusher.rb
@@ -28,9 +28,9 @@ module Gemstash
 
     def serve
       check_auth
-      save_to_database
       store_gem
       store_gemspec
+      save_to_database
       invalidate_cache
     end
 
--- a/lib/gemstash/gem_pusher.rb
+++ b/lib/gemstash/gem_pusher.rb
@@ -54,8 +54,7 @@ module Gemstash
 
     def store_gem
       resource_exist = storage.resource(full_name).exist?
-      resource_is_indexed =
-        storage.resource(full_name).properties[:indexed] if resource_exist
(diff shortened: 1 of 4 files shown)


bronzdoc:
A spec in jruby is failing but i can't reproduce the error locally...

Running `bundle rake rspec`
```shell
.....
Finished in 1 minute 46.58 seconds (files took 6.38 seconds to load)
257 examples, 0 failures, 3 pending
```

```shell
$ rbenv versions
  system
  2.2.10
  2.3.7
  2.4.1
  2.4.4
  2.5.0
  2.5.1
* jruby-9.2.0.0 (set by /home/bronzdoc/projects/ruby/gemstash/.ruby-version)
```

The build fails in ruby-2.4.1 and ruby and in ruby-2.3.4 got fixed by restarting the builds.

bbuchalter:
> An integration spec in jruby is failing but i can't reproduce the error locally...
> The build failing in ruby-2.4.1 and ruby and in ruby-2.3.4 got fixed by restarting the builds.

https://travis-ci.org/bundler/gemstash/jobs/440896430#L513: `The job exceeded the maximum time limit for jobs, and has been terminated.`

Maybe we can retry the jruby build too?

bronzdoc:
@bbuchalter 
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

^ That is me running the build in `debug` mode, travis allows you to get inside the build container to debug, that is not the actual failure.

I will work on it later tonight

lib/gemstash/gem_pusher.rb
```diff
@@ -53,6 +53,13 @@ def check_auth
     end
 
     def store_gem
+      resource_exist = storage.resource(full_name).exist?
+      resource_is_indexed = storage.resource(full_name).properties[:indexed] if resource_exist
+
+      if resource_exist && resource_is_indexed
```
bbuchalter: This guard is mirrored in [`save_to_database`](https://github.com/bundler/gemstash/blob/f14894610a1cbef21d6f595ba3b332e45baa5f32/lib/gemstash/gem_pusher.rb#L79). Perhaps it should be extracted to a private method?

Also, there is a [`YankedVersionError` guard](https://github.com/bundler/gemstash/blob/f14894610a1cbef21d6f595ba3b332e45baa5f32/lib/gemstash/gem_pusher.rb#L80) in `save_to_database`. Should that _also_ be here?

bronzdoc:
For the record... the build failure is not related to this PR, the failure is happening in `master`
 
The specs pass if change the ruby version to `jruby-9.1.17.0` inside the build container.

lib/gemstash/gem_pusher.rb
bronzdoc: WIll do a followup PR if applies, thanks

bronzdoc:
@bundlerbot r+

Issue #99, opened 2016-06-01
Raised exceptions should include an appropriate HTTP status code and respond better than a 500 error


smellsblue:
`Gemstash::GemPusher::ExistingVersionError` raises and triggers a raw 500 Sinatra error. It should instead have a more appropriate HTTP status code with a nicer response. This goes for all Gemstash exceptions that get raised.

bronzdoc:
closed by https://github.com/bundler/gemstash/pull/190

PR #192, opened 2018-10-27
Fix #188 – Prevent overriding published gems

### LINKED ISSUE #188: Gemstash allows overriding published gems

mzruya:
This PR attempts to fix issue #188 (and the relevant parts of #99).
It does so by re-ordering the storage sequence in the `gem_pusher.rb` file.
Since the database should be the source of truth, checking it first ensures we raise an exception before any on disk side-effects occur.

Longer term, I think it makes sense to use a different storage path on disk (UUID's?) and perhaps even store the checksum in the database. this would allow ensuring the files have not been tampered with.

I also decided to take a stab at part of #99 and ensure that the exception raised is caught and properly handled by the controller.
I opted for 422 as the error code.

### DIFF
--- a/lib/gemstash/gem_pusher.rb
+++ b/lib/gemstash/gem_pusher.rb
@@ -28,9 +28,9 @@ module Gemstash
 
     def serve
       check_auth
+      save_to_database
       store_gem
       store_gemspec
-      save_to_database
       invalidate_cache
     end
 
--- a/lib/gemstash/web.rb
+++ b/lib/gemstash/web.rb
@@ -1,7 +1,6 @@
 require "sinatra/base"
 require "json"
 require "gemstash"
-require "pry"
 module Gemstash
   #:nodoc:
   class Web < Sinatra::Base
--- a/lib/gemstash/web.rb
+++ b/lib/gemstash/web.rb
@@ -1,6 +1,7 @@
 require "sinatra/base"
 require "json"
 require "gemstash"
+
 module Gemstash
   #:nodoc:
   class Web < Sinatra::Base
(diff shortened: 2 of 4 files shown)


mzruya:
@bronzdoc, I only noticed your change (#192) after creating this PR :-)
This change also includes controller exception handling. feel free to cherry-pick some of it if you see fit.
Thanks for working on a fix.

bronzdoc:
Thanks @mzruya will do!

bronzdoc:
closing in favor of https://github.com/bundler/gemstash/pull/190

PR #193, opened 2018-10-29
Fix jruby builds

bronzdoc:
Install jruby manually in a generic language image, since we can't modify the `/home/travis/.rvm/gemsets/jruby/global.gems` in what travis gives us.

travis `home/travis/.rvm/gemsets/jruby/global.gems`  has in it `jruby-openssl` wich makes rvm install a new version of `jruby-openssl` whenever we install jruby, making the test suite run with that `jruby-openssl` version instead of the one that comes with jruby itself an break.


bronzdoc:
@bundlerbot r+

PR #195, opened 2018-10-31
GEMSTASH-194 Support for FIPS Mode

### LINKED ISSUE #194: Support hosts in FIPS mode
Currently, gemstash does not run on systems that are configured in FIPS mode. There are a couple spots that utilize MD5 digests that need to be replaced for valid operation. Results in the following logs when running the server and timeouts for too many connection resets in fetching gems:

```
md5_dgst.c(82): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
[2018-10-31 13:30:47 +0000] - INFO - [16128] - Worker 0 (pid: 16242) booted, phase: 0
md5_dgst.c(82): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
[2018-10-31 13:30:52 +0000] - INFO - [16128] - Worker 0 (pid: 16254) booted, phase: 0
md5_dgst.c(82): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
[2018-10-31 13:30:57 +0000] - INFO - [16128] - Worker 0 (pid: 16266) booted, phase: 0
md5_dgst.c(82): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
[2018-10-31 13:31:02 +0000] - INFO - [16128] - Worker 0 (pid: 16276) booted, phase: 0
```

ayohrling:
This updates the digest usage from MD5 to SHA256 to support hosts that are
configured in FIPS mode.

Fixes #194

### DIFF
--- a/lib/gemstash/storage.rb
+++ b/lib/gemstash/storage.rb
@@ -114,7 +114,7 @@ module Gemstash
       trie_parents = safe_name[0...3].downcase.split("")
       # The digest is included in case the name differs only by case
       # Some file systems are case insensitive, so such collisions will be a problem
-      digest = Digest::MD5.hexdigest(@name)
+      digest = Digest::SHA256.hexdigest(@name)
       child_folder = "#{safe_name}-#{digest}"
       @folder = File.join(@base_path, *trie_parents, child_folder)
       @properties = nil
--- a/lib/gemstash/upstream.rb
+++ b/lib/gemstash/upstream.rb
@@ -40,7 +40,7 @@ module Gemstash
   private
 
     def hash
-      Digest::MD5.hexdigest(to_s)
+      Digest::SHA256.hexdigest(to_s)
     end
 
     #:nodoc:


olleolleolle:
If I understand this correctly, this change modifies the directory names for Storage's "Resources" (aka uploaded things).

That would be a breaking change, right?

ayohrling:
Yes, it changes the hashing mechanism used in the storage structure to a FIPS-compliant cipher. It will break any cached resources from older versions. Upgraded systems would need to re-cache. A conversion really wouldn't be possible, because we'd be guessing at rolling back from `safe_name` to `name` with regards to casing.

bronzdoc:
Maybe add this as a flag at the moment and make it the default in a major version.

olleolleolle:
To suggest a flag name, I read a bit about the "FIPS mode" concept. 

I like the MS registry key name best: `FipsAlgorithmPolicy`.

It describes what we do to Gemstash when this new option is set. We choose crypto in FIPS-compliant ways.

As an option name `--use-fips-algorithm-policy` (default: `false`).

What are some better flag names?

bronzdoc:
@olleolleolle  what about `--fips` and have what it does in the description?

olleolleolle:
I added a PR on top of the @ayohrling one.

Issue #203, opened 2018-11-25
Error: Could not find command "readme"


zpdcodes:
An error raised when run `gemstash` or `gemstash -h`:

    Could not find command "readme"

However the `gemstash start` works fine.

olleolleolle:
Steps to reproduce:

Started with `gem install gemstash`:

```
$ gem install gemstash
$ gemstash --version
Gemstash version 2.0.0
$ gemstash --help
Could not find command "readme".
```

olleolleolle:
Hm, could it be that the distributed .txt files are not found?

```ruby
        page = File.expand_path("../man/gemstash-#{command}.#{section}", __FILE__)
```

could be turned into

```ruby
        page = File.expand_path("../man/gemstash-#{command}.#{section}.txt", __FILE__)
```

gepser:
Hi, I investigated this issue and here is what I found:

**Status**:
In the latest release (`2.0.0`) the task `doc` (`rake doc`) didn't run (or ran with errors) so the docs are missing. Gemstash works just fine but you can't see any docs from the cli.

**Possible solution**:
Make a new release running `rake doc` before actually running `rake release`. I guess you can make one task to depend on the other.

**Tests**:
I actually tested all of this on my machine and inside a docker container and it works fine, docs are not there if you don't run `rake doc` and docs are there when you actually run `rake doc`.

**Notes**:
- On my tests I got this error: `You need to install pandoc to generate documentation` so you are gonna need `pandoc` installed on the machine to do the release.
- Part of the release process is to publish the gem on rubygems and I guess I would need special permissions for that. That's why I am not making the release myself.
- Anyhow, I just wanted to document my findings before trying to solve another issue.

 I hope this helps.

bronzdoc:
This is fixed in `Gemtash 2.1.0` thanks all!

PR #204, opened 2018-11-25
Rake task doc: drop fix_github_line_breaks filter; change to gfm Pandoc formatter

olleolleolle:
This PR tries to run the Rake `doc` task without Pandoc errors or warnings, to repair documentation generation.

The `doc` Rake task calls a list of Pandoc Filters. This PR removes one of those Filters.

See #203

# Benefits

This change made the files at least _render_, and the content line-breaks is not created using double line-breaks.

The change has 3 commits. The last commit is the "regenerate the docs with this". It's included, so that any troublesome renderings are visible here.

# Changes

Some things have **changed** with us using `gfm` as a pandoc Markdown formatter.

- `> $ gemstash setup --redo` the `--` becomes a n-dash or something (the `smart` extension of Pandoc is doing this). _Work-around_: I added code formatting back-ticks around the shell example
- CHANGE in the output Markdown content: [ATX headers](https://pandoc.org/MANUAL.html#headers) are now the rendered default. (`#` instead of `-----`). [All about Pandoc Headers](https://pandoc.org/MANUAL.html#headers).

> `--atx-headers` 
> Use ATX-style headers in Markdown output. The default is to use setext-style headers for levels 1-2, and then ATX headers. (**Note**: for `gfm` output, ATX headers are always used.)

# Questions for reviewer

- Does this change any of the other filters? Their meaning, their input/output?
- Am I missing some important change in output?
- Does this work on your computer?

<details>

Versions used

```
$ pandoc --version
pandoc 2.4
Compiled with pandoc-types 1.17.5.4, texmath 0.11.1.2, skylighting 0.7.4
Default user data directory: /Users/olle/.pandoc
Copyright (C) 2006-2018 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.

$ ruby -v
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin17]
```

</details>


indirect:
I don't think I can answer any of those questions, but I am in favor of docs that actually generate... on Travis would be best, so we can potentially auto-release when a Travis build passes for a new tag?

bronzdoc:
> Does this change any of the other filters? Their meaning, their input/output?
I don't think so since the only filter you are changing is `markdown_github`, unless i'm missing something?

> Am I missing some important change in output?
I'm not 100% sure about that, since i'm not a pandoc expert, but markdown_github waa only used to for json output and i don't see any other output being used with `markdown_github` in the codebase.

Does this work on your computer?
It removes the warnings as you mention, so yes it works, but it does not solve the problem of https://github.com/bundler/gemstash/issues/203 (Error: Could not find command "readme")

I'll say lets merge this @olleolleolle  since is replacing a deprecated pandoc filter for what pandoc recomends we change it toif we encounter any other problems regarding pandoc we can create another PR and fix them along the way.

Just please squash your commits into meaningful chunks or remove the "WIP" of the commit message, thanks for this!

bronzdoc:
@olleolleolle please use @bundletbot to merge PR's thanks!

olleolleolle:
Ouch, _yes_!

PR #210, opened 2018-12-10
Avoid Rubocop warnings

olleolleolle:
This PR avoids the warnings output from Rubocop

```
..................................spec/support/simple_server.rb: Style/MethodName has the wrong namespace - should be Naming
..spec/support/slow_simple_server.rb: Style/MethodName has the wrong namespace - should be Naming
.................................................................```


bronzdoc:
@bundlerbot r+

PR #240, opened 2019-06-19
Fix broken links

Aliciawyse:
Three of the documentation links in the section titled "Under the Hood" were broken. I prefixed each path with `docs/` to fix the broken link.


hsbt:
Ah, sorry. I forgot to use bundlerbot.

PR #242, opened 2019-06-24
Add backends for Gemstash storage of private gems

gayashanbc:
## Description
This pull request reflect the changes related to the GSoC project, "[Add backends for Gemstash storage of private gems](https://summerofcode.withgoogle.com/projects/#5922066392416256)".

## Blog Posts
1. [Introduction & Implementing Support for The ‘backend’ Property](https://medium.com/@gayashanbc/gsoc-2019-episode-1-introduction-implementing-support-for-the-backend-property-931b71f13f99)
2. [Implementing Trigger Function for AWS S3 Backend](https://medium.com/@gayashanbc/gsoc-2019-episode-2-implementing-trigger-function-for-aws-s3-backend-2f5c8b5eb7f7)

### DIFF
--- a/lib/gemstash/cli/setup.rb
+++ b/lib/gemstash/cli/setup.rb
@@ -126,12 +126,11 @@ module Gemstash
 
       def ask_backend
         say_current_config(:backend, "Current backend")
-        previous_backend = @config[:backend]
         @config[:backend] = ask_with_default("What backend?", %w[local s3], "local")
-        check_aws_authorization(previous_backend) unless @config[:backend] == "local"
+        check_aws_authorization() unless @config[:backend] == "local"
       end
 
-      def check_aws_authorization(previous_backend)
+      def check_aws_authorization
         begin
           @cli.say "Checking authorization to Aws"
           s3 = Aws::S3::Resource.new()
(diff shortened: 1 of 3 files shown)


lib/gemstash/cli/setup.rb
```diff
@@ -26,6 +26,7 @@ def run
         ask_database
         ask_protected_fetch
         ask_timeout
+        ask_bachend
```
olleolleolle:
```suggestion
        ask_backend
```

lib/gemstash/cli/setup.rb
```diff
@@ -122,6 +123,11 @@ def ask_timeout
         @config[:fetch_timeout] = timeout.to_i
       end
 
+      def ask_bachend
```
olleolleolle:
```suggestion
      def ask_backend
```

lib/gemstash/cli/setup.rb
```diff
@@ -122,6 +123,11 @@ def ask_timeout
...
+        say_current_config(:backend, "Current backend")
+        @config[:backend] = ask_with_default("What backend?", %w[local s3], "local")
```
olleolleolle: Perhaps: Only keep `local` in the list of available backends, for now?

indirect:
This is a good start! @gayashanbc, let's plan to continue adding commits to this PR, and we can give feedback and code review when you push or at least once a week. Once this PR has a more complete version of the backends feature, we can merge it.

lib/gemstash/cli/setup.rb
gayashanbc: Fixed

lib/gemstash/cli/setup.rb
gayashanbc: Fixed

lib/gemstash/cli/setup.rb
gayashanbc: I'm working on implementing the trigger actions for each backend including s3.
As a part of the process authorization to s3 is implemented.
So is it OK to keep both 'local' and 's3'?

gayashanbc:
aws-sdk is added to the Gemfile.
I have tested and verified the ability to verify the SDK using either of the following options.

1.  Default AWS credentials file (e.g. on Unix-like systems this file is located at `~/.aws/credentials`)
2. Using environmental the environmental variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`

Currently, I'm working on displaying an error message when the SDK is not authorized using either of the above methods and displaying a link to official AWS page where it shows how the AWS credentials can be configured on the system. This will be implemented in the trigger action, when the backend is selected as 's3'.

lib/gemstash/cli/setup.rb
olleolleolle: Right, I take it that this PR will contain the whole change, including support for both of the named backends.

Totally OK to keep both.

Thanks for sharing your plans, it's a good feature.

gayashanbc:
Above two commits implement the authorization trigger for s3 backend and the initial version of ask_s3_metadata method which checks if the given bucket_name already exists in AWS S3.

The next step is to create an S3 bucket using the given bucket_name if it does not exist already.

gayashanbc:
Update: Above commit implements support for creating a new S3 bucket when the specified bucket is not owned by the user. It will also display an error when specified bucket_name is already taken by another user.

bronzdoc:
closing this since a s3 backend was added in https://github.com/rubygems/gemstash/pull/270 thanks!

PR #244, opened 2019-07-08
Improve webspec tests

Aliciawyse:
Some of the tests did not check against an explicit expectation. This updates the test to use rspec matchers to clearly express the intention of the test.


bronzdoc:
@bundlerbot r+

PR #248, opened 2019-09-04
Add puma server pidfile configuration option

krists:
It would be useful to allow setting up custom puma pidfile locations. For example you might want to set it to `/var/run/gemstash-puma.pid` so that the file does not persist in case of system is not shut down correctly.


bronzdoc:
Please add a test for this, thanks!

krists:
@bronzdoc I have added couple of test cases to spec/gemstash/env_spec.rb file where similar options has been tested.

bronzdoc:
@bundlerbot r+

PR #250, opened 2019-11-29
Officially drop support for older rubies

deivid-rodriguez:
In https://github.com/rubygems/gemstash/pull/198 it was discussed to drop support for old rubies, and then testing against those rubies stopped at https://github.com/rubygems/gemstash/pull/201. But support was not actually dropped in the sense that the gem still installs in old rubies. It might be already broken though since they haven't been tested for a while.

Anyways, this PR should prevent installation on those no longer supported rubies.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,3 +8,4 @@ AllCops:
     - vendor/**/*
     - gemfiles/**/*
   DisplayCopNames: true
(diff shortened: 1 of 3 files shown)


bronzdoc:
Thanks @deivid-rodriguez 

@bundlerbot r+

PR #260, opened 2020-02-26
Update docs

bronzdoc:
# Description:
Updates README.md
Updates man 
Update docs 
Updates image links generation
Fixes broken doc links generation
______________

# Tasks:

-  Write tests

### DIFF
--- a/rake/doc/fix_links_for_format.rb
+++ b/rake/doc/fix_links_for_format.rb
@@ -56,8 +56,7 @@ class DocLinkUrl
 
   def relative_path
     current = Pathname.new(current_path)
-    file = Pathname.new(path).join(filename)
-    file.relative_path_from(current).to_s
+    Pathname.new(path).join(filename)
   end
 
   def heading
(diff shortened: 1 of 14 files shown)


rake/doc/fix_links_for_format.rb
```diff
@@ -55,9 +55,7 @@ def format_extension
   end
 
   def relative_path
-    current = Pathname.new(current_path)
-    file = Pathname.new(path).join(filename)
-    file.relative_path_from(current).to_s
+    Pathname.new(path).join(filename)
```
bronzdoc: @olleolleolle do you see something wrong with doing this change?

olleolleolle: Looking at the code, and imagining: 

If we were using very relative paths that needed resolving, then the presence of "current path" would perhaps be important. 

Or, hm, is it there in order to _create_ very terse relative paths?

I think the new code reads easier, but perhaps does something else, which wouldn't be a bad thing, in this case, I guess. 

Ok, speculation ends here. 

I don't know, but I'm curious!

bronzdoc:
Without this `rake doc` will generate wrong links in the `README`

bronzdoc:
@bundlerbot r+

PR #264, opened 2020-03-19
Abstract local storage

nicolas-brousse:
# Description:

This PR intent to abstract current local storage, and make easier to add custom storage (like S3 storage).

I didn't finish test yet, but I may work on it if you're interested on this feature.

Related to #107, #143, #159, #242

______________

# Tasks:

-  Describe the problem / feature
-  Write tests

I will abide by the [code of conduct](https://github.com/gemstash/gemstash/blob/master/CODE_OF_CONDUCT.md).

PR #267, opened 2020-06-05
S3 Implementation 

duyquangnguyenhac:
**#**Description**:**
With consideration of the past S3-related commits, I opened this pull request in order to work on it full time for this year's GSOC program. As of now, the work is not done, but I've created a mock S3 storage client based on the current storage implementation with all of the same methods, but with S3 implementation. This is so I could understand the concept of the task at hand, and make sure I'm heading towards the right direction.

The S3 class created is mostly functional with several bugs to be fixed, there are several RSpec tests I've made to ensure its functionality. If you would like to run those RSpec tests, you can by providing your own credentials manually in the default config section. This is the way to do it as of now because I haven't done any changes to the CLI. Eventually, the tests will move away from VCR because VCR testings are not exactly optimal. The class can also successfully run with gem_pusher tests with small fixes to the expected parameters.

**#Bugs at hand**
- The @content uploaded to the S3 servers are encoded in UTF-8 which is different from the binary encoding on the filesystem files which causes the tests on modules such as the gem_pusher to fail.
-  Issues accessing the gemstash_env variables for S3 storage.

**#Features to be implemented:**
- Implementation into the CLI
- In-depth tests for the classes
- Abstractualized storage class as presented #264
- Stub-testing instead of VCR


lib/gemstash/S3.rb
```diff
@@ -0,0 +1,247 @@
...
+require "yaml"
+require "aws-sdk-s3"
+require "pathname"
+
+module Gemstash
+  class S3
+    extend Gemstash::Env::Helper
+    attr_reader :folder, :client, :bucket
+
+    VERSION = 1
```
bronzdoc, line 13: Maybe call this `STORAGE_VERSION`?

lib/gemstash/S3.rb
```diff
@@ -0,0 +1,247 @@
...
+    class VersionTooNew < StandardError
```
bronzdoc, line 14: Maybe call it `VersionTooNewError` ?

lib/gemstash/S3.rb
```diff
@@ -0,0 +1,247 @@
...
+      def initialize(folder, version)
+        super("Gemstash storage version #{Gemstash::S3::VERSION} does " \
```
bronzdoc, line 16: Can you explain what this error means?

spec/gemstash/S3_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+require "spec_helper"
+require "yaml"
+
+RSpec.describe Gemstash::S3 do
+  let(:storage) { Gemstash::S3.new("gemstash/S3_storage_tests",Gemstash::Env.current).for("private").for("gems") }
+  before do
+    @folder = Dir.mktmpdir
+  end
+  after(:all) do
+    VCR.use_cassette('batch_delete objects') do
```
bronzdoc, line 12: Try to be consistent with the style, use double quotes

Gemfile
```diff
@@ -3,3 +3,7 @@
 source "https://rubygems.org"
 
 gemspec
+
+gem 'aws-sdk-s3', '~> 1'
```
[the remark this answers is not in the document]
bronzdoc: Also I think this should be now a runtime dependency

lib/gemstash/S3.rb
```diff
@@ -0,0 +1,247 @@
...
+        memo
+      end[:result]
+    end
+
+    def delete(key)
+      return self unless exist?(key)
+
+      begin
+        @S3resource.object(content_filename(key)).delete
+      rescue StandardError => e
```
bronzdoc, line 152: I think we want to catch the specific error that `@S3resource.object(content_filename(key)).delete` can cause so we don't so we can rescue properly

lib/gemstash/S3.rb
```diff
@@ -0,0 +1,247 @@
...
+        log_error "Failed to delete stored content at #{content_filename(key)}", e, level: :warn
+      end
+
+      begin
+        @S3resource.object(properties_filename).delete unless content?
+      rescue StandardError => e
```
bronzdoc, line 158: Same as above

lib/gemstash/S3.rb
```diff
@@ -0,0 +1,247 @@
...
+
+      entries = @S3resource.objects(prefix: (@folder)).collect().reject { |object| object.content_length == 0 || object.key == "properties.yaml" }
+      !entries.empty?
+    end
+
+    def load_properties(force = false)
+      return if @properties && !force
+      return unless @S3resource.object(properties_filename).exists?
+
+      properties_file = @S3resource.object(properties_filename).get.body
```
bronzdoc, line 180: maybe rename this to `properties_content` since is not a file but the `body`

lib/gemstash/S3.rb
```diff
@@ -0,0 +1,247 @@
...
+    def read_file(filename)
+      @S3resource.object(filename).get().body.read
+    end
+
+    def properties_filename
+      File.join(@folder, "properties.yaml")
+    end
+
+    def save_file(filename)
+      content = yield
```
bronzdoc, line 237: why we need to yield the content and to pass it as an argument?

lib/gemstash/S3.rb
```diff
@@ -0,0 +1,247 @@
...
+    def load_properties(force = false)
+      return if @properties && !force
+      return unless @S3resource.object(properties_filename).exists?
+
+      properties_file = @S3resource.object(properties_filename).get.body
+      @properties = YAML.load(properties_file) || {}
+      check_resource_version
+    end
+
+    def save_properties(props)
```
bronzdoc, line 185: we can set the default of props as `props  = {}`

lib/gemstash/S3.rb
duyquangnguyenhac: If you take a look at the initialization at line 23, it is used to make sure that when the S3 Storage class is originally initialized, the resource version is checked. If root = false, or if the initialization isn't the original initialization, but merely adding on to a previous initialization, then it is not checked. I'm unsure how the addition of a root variable helps, but I think it improves performance so that you don't have to check the version if it isn't the root initialization.

lib/gemstash/S3.rb
duyquangnguyenhac: In hindsight, this method yield is certainly a bit unnecessary. I leave it as is because it was already implemented in the previous storage class so I didn't want to cause unexpected errors. At some point, I will refactor the entire storage class to make it simpler, and more straightforward.

lib/gemstash/S3.rb
duyquangnguyenhac: Fixed in commit #bfa012a81

lib/gemstash/S3.rb
duyquangnguyenhac: resolved above ^

Gemfile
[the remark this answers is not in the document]
duyquangnguyenhac: Addressed in 0944f59bf5be836264f59970485ff21f6e825a83

lib/gemstash/S3.rb
duyquangnguyenhac: Addressed in 8c8713bf5f9b37e8a2e8e889a3f614e25141b9de

lib/gemstash.rb
```diff
@@ -27,4 +27,6 @@ module Gemstash
   autoload :Web,                 "gemstash/web"
   autoload :WebError,            "gemstash/http_client"
   autoload :VERSION,             "gemstash/version"
+  autoload :S3,                  "gemstash/S3"
```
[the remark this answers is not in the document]
duyquangnguyenhac: Because s3 class and s3 resource class have the same location in s3_storage.rb
Refractored in 8c8713bf5f9b37e8a2e8e889a3f614e25141b9de

PR #270, opened 2020-07-01
Gemstash S3 Backend Functionality

duyquangnguyenhac:
# Description:
Re-open previously closed PR.

With consideration of the past S3-related commits, I opened this pull request in order to work on it full time for this year's GSOC program. As of now, the work is not done, but I've created a mock S3 storage client based on the current storage implementation with all of the same methods, but with S3 implementation. This is so I could understand the concept of the task at hand, and make sure I'm heading towards the right direction.

The S3 class created is pretty much entirely functional. If you would like to run those RSpec tests, you can by providing your own credentials manually in the config.yml file located in the [~/.gemstash] directory. 

A functional S3 implementation for gemstash, which introduce the command flag -s3 to gemstash setup command. To run the gemstash module. 
```ruby
gemstash setup --redo --s3
```
Then provide the appropriate credentials.

______________

# Notes:
- Method `Gemstash::S3#delete_with_prefix` is not ideal, but is needed to clear the temporary files made for testing purposes.
- Working on new CLI test.


Gemfile
```diff
@@ -3,3 +3,7 @@
 source "https://rubygems.org"
 
 gemspec
+
+gem "aws-sdk-s3", "~> 1.67"
+gem "vcr"
+gem "webmock"
```
duyquangnguyenhac, line 9: @bronzdoc Where should I move these gems to?

lib/gemstash/storage_services/s3_storage.rb
```diff
@@ -0,0 +1,289 @@
...
+      @object_name = @folder
+    end
+
+    def check_credentials?
+      @client.get_bucket_location(bucket: @bucket_name).location_constraint == gemstash_env.config[:region]
+    end
+
+    def delete_with_prefix(prefix = @folder)
+      Aws::S3::Resource.new(client: @client).bucket(@bucket_name).objects(prefix: prefix).batch_delete!
+    end
```
duyquangnguyenhac, line 47: @bronzdoc This is the method I was talking about. It is used in https://github.com/duyquangnguyenhac/gemstash/blob/8d51e62440c5b5756971bbc8507ada54c5904e42/spec/gemstash/s3_storage_spec.rb#L23
It is not wise to have such delete method, but I use it to clear the storage after I run the VCR tests. How should I fix this ?

Gemfile
bronzdoc: gemspec

lib/gemstash/configuration.rb
```diff
@@ -20,7 +20,13 @@ class Configuration
       puma_threads: 16,
       puma_workers: 1,
       cache_expiration: 30 * 60,
-      cache_max_size: 500
+      cache_max_size: 500,
+      storage_adapter: "local",
```
bronzdoc, line 24: All this new config will need documentation here https://github.com/rubygems/gemstash/blob/master/docs/gemstash-customize.7.md

lib/gemstash/gem_pusher.rb
```diff
@@ -3,7 +3,6 @@
 require "gemstash"
 require "rubygems/package"
 require "stringio"
-
```
bronzdoc, line 6: ?

lib/gemstash/storage.rb
```diff
@@ -7,343 +7,65 @@
...
-      content = yield
-      gemstash_env.atomic_write(filename) {|f| f.write(content) }
-    end
-
-    def read_file(filename)
-      File.open(filename, "rb", &:read)
+    def self.metadata
+      storage_service = gemstash_env.config[:storage_adapter]
+      case storage_service
+      when "local"
```
bronzdoc, line 54: We are doing this check of "local" and "s3" in a few places, maybe we can put it in a method

duyquangnguyenhac: Ideally, this would be solved by #264, it should handle all storage_service selection choices in one dedicated place.

lib/gemstash/configuration.rb
duyquangnguyenhac: Added in c072f9d.

gemstash.gemspec
```diff
@@ -32,6 +32,7 @@ you push your own private gems as well."
   spec.required_ruby_version = ">= 2.4"
 
   spec.add_runtime_dependency "activesupport", ">= 4.2", "< 6"
+  spec.add_runtime_dependency "aws-sdk-s3", "~> 1.67"
```
indirect, line 35: Let's make this dependency work the same way as redis or mysql down on lines 45-50, where it is not a hard requirement. Instead, document that anyone wanting to use gemstash with S3 will need to install the aws-sdk-s3 gem as wall. That way gemstash users who don't want to use s3 don't have to install the s3 gem.

lib/gemstash/storage_services/s3_storage.rb
indirect: You could add this method to Gemstash::S3 only inside the tests. In one of the support files loaded by the tests, something like this:

```ruby
module Gemstash
  class S3
    def delete_with_prefix(prefix = @folder)
       Aws::S3::Resource.new(client: @client).bucket(@bucket_name).objects(prefix: prefix).batch_delete!
     end
  end
end
```

indirect:
This is looking pretty good! I'm +1 to merge it once the remaining comments from myself and @bronzdoc are addressed, and the tests are passing.

lib/gemstash/gem_pusher.rb
duyquangnguyenhac: This is probably just a mistake

lib/gemstash/storage_services/s3_storage.rb
duyquangnguyenhac: Thank you so much for your suggestion. I did move that method into a helper method that's only used during tests!
It is fixed in this commit: e7b7336

gemstash.gemspec
duyquangnguyenhac: Sorry for the late reply, but if I'm understanding this correctly, if this is added, the user will need to manually edit gemstash's gemspec to include the s3 gem in-order for this feature to work right? I'm just making sure it is added as intended.

bronzdoc: @duyquangnguyenhac nope. The user does't have to do that they will just need to install the gem manually.

Add it just as a development dependency, just like redis.

indirect: Yeah, no editing of gemspec. Either they run “gem install”, or they use a Gemfile that has both gemstash and aws-sdk-s3 in it.

If you want to get fancy, you can rescue errors while trying to require and print a message that the user needs to install aws-sdk-s3 and try again. 🙂

duyquangnguyenhac: Done 👍
Moved the gem into development dependency and adds a custom error message when requiring load error occurs.
Added in d688581.

duyquangnguyenhac:
Just an update, this PR is still good so far. It is not passing the tests because of an error unrelated to this branch and affects even the master branch!

bronzdoc:
@bundlerbot try

bronzdoc:
Test failures are not related to this PR

Thanks so much for your hard work @duyquangnguyenhac!

kaycoinsUSA (commented):
Great work,Great Team!

muriloime:
Why was this reverted?

olleolleolle:
@muriloime See #293. Hope this helps!

PR #300, opened 2021-07-30
Add info command

bronzdoc:
# Description:

Adds back info command without the failing spec


rsanheim:
This is pretty old, but wondering if it could get merged? I'm having a difficult time figuring out the current state of gemstash with some different configs I'm trying - this would be useful.

indirect (approved):
Going to defer to @olleolleolle on the actual merge, but I'm 👍🏻 on this.

spec/gemstash/cli/info_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    Gemstash::Env.current = @test_env
+  end
+
+  let(:cli_options) do
+    {
+      config_file: File.join(TEST_BASE_PATH, "info_spec_config.yml")
+    }
+  end
+
+  it "Prints the current current Gemstash configuration" do
```
olleolleolle, line 49:
```suggestion
  it "prints the current current Gemstash configuration" do
```

Minor: Perhaps the `it` starts on a lower-case letter.

olleolleolle (approved):
So nice, I am in favor and thanks, @bronzdoc!

spec/gemstash/cli/info_spec.rb
olleolleolle: ("We can fix it in post" as the film-makers say.)

PR #301, opened 2021-09-26
Update gemstash-customize.7.md

simi:
fix broken link

It doesn't work while browsing github. But I'm not sure if this is not introducing any problems for generated man pages.


olleolleolle:
https://github.com/rubygems/gemstash/pull/260 has a repair to exactly what you suspected, so I _guess_ we can't do it just like this. Some other tricky way is needed, where both the generated pages AND the navigatable pages on GitHub are in tandem.

simi:
ok, thanks for the info

PR #309, opened 2021-11-08
build: Try to use a Bundler group for lint

olleolleolle:
# Description:

RuboCop in a Bundler group, and use caching.

Hm, this does not skip the installation of the other stuff, the group, as used here - BUT the bundler-cache _works_.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,14 +6,14 @@ jobs:
   rubocop:
     name: Rubocop
     runs-on: ubuntu-18.04
+    env:
+      BUNDLE_WITH: "linting"
     steps:
     - uses: actions/checkout@v2
     - uses: ruby/setup-ruby@v1
       with:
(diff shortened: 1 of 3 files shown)


tonytonyjan:
I think we should put development dependencies in either Gemfile or gemspec to make it clear.
Until today, Ruby community still doesn't have an convention for whether put them in gemspec by `add_development_dependency` or in `Gemfile`.

Putting some in `Gemfile` and some in Gemspec file might confuse people in my opnion.

refs:

1. https://github.com/rubygems/rubygems/issues/1104
2. https://stackoverflow.com/questions/18871541/what-is-the-purpose-of-test-files-configuration-in-a-gemspec

olleolleolle:
Yes, I agree, and I would prefer the Gemfile  to be the home of all of them.

olleolleolle:
@tonytonyjan I will make a PR real quick.

PR #312, opened 2021-11-11
feat: support Ruby 3

### LINKED ISSUE #285: incombatible with Ruby-3 / later gem versions (wrong number of arguments)
Hi, 

just to make the report reproducible, I'm using a docker file. This dockerfile works  (with ruby-2.7):

FROM ruby:2-alpine

RUN apk add --update --upgrade --no-cache --virtual .build-dependencies gcc musl-dev make sqlite-dev && \
    gem install gemstash && \
    apk del .build-dependencies && \
    rm -rf /var/cache/apk/*

CMD ["gemstash","start","--no-daemonize"]

But when replacing the FROM image with ruby:alpine (i.e. using ruby 3) it still compiles,yet the process results in 

/usr/local/bundle/gems/gemstash-2.1.0/lib/gemstash/web.rb:10:in `initialize': wrong number of arguments (given 1, expected 0) (ArgumentError)
    from /usr/local/bundle/gems/sinatra-2.1.0/lib/sinatra/base.rb:1527:in `new'
   [...]

regards

tonytonyjan:
fix #285

### DIFF
--- a/.github/workflows/gemstash-ci.yml
+++ b/.github/workflows/gemstash-ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.4, 2.5, 2.6, 2.7, jruby-9.2]
+        ruby: [2.4, 2.5, 2.6, 2.7, "3.0", jruby-9.2]
     steps:
       - uses: actions/checkout@v2
       - name: Setup ruby
--- a/lib/gemstash/web.rb
+++ b/lib/gemstash/web.rb
@@ -7,9 +7,11 @@ require "gemstash"
 module Gemstash
   #:nodoc:
   class Web < Sinatra::Base
-    def initialize(gemstash_env: nil, http_client_builder: nil)
-      @gemstash_env = gemstash_env || Gemstash::Env.new
-      @http_client_builder = http_client_builder || Gemstash::HTTPClient
+    ruby2_keywords def initialize(options = {})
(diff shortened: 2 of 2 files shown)


.github/workflows/gemstash-ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.4, 2.5, 2.6, 2.7, jruby-9.2]
+        ruby: [2.4, 2.5, 2.6, 2.7, 3.0, jruby-9.2]
```
olleolleolle, line 11: Oh, this needs a quote

```suggestion
        ruby: [2.4, 2.5, 2.6, 2.7, "3.0", jruby-9.2]
```

This is a very small change, but which avoids a Float-to-String loss of characters.

An example (not from this project) of what it can look like before this change, in this picture:

bild

Read more details, if you like: https://github.com/actions/runner/issues/849

olleolleolle: bild

Here is the output from this run, of this PR.

olleolleolle: By quoting, we'd see "3.0", there.

olleolleolle (approved):
Thanks @tonytonyjan!

.github/workflows/gemstash-ci.yml
tonytonyjan: @olleolleolle Thanks for the reminder  :)

Issue #247, opened 2019-08-29
Mirroring for repos requiring authentication


viraptor:
I couldn't find a description of how I could run gemstack mirroring for a specific rubygems repo which requires authentication.

I've got the credentials for the repo configured in `~/.bundle/config`, but if I specify `BUNDLE_MIRROR__....` for it as well, the download fails. Am I missing something trivial?

tjwp:
I just figured out how to get this to work.

The credentials from `~/.bundle/config` don't get used. Maybe that is due to bundler not passing them along to the mirror when using the simpler approach described here? https://github.com/rubygems/gemstash/blob/master/docs/gemstash-mirror.7.md

The solution that I found was to include the `/upstream` path when specifying the gemstash mirror and include the credentials in the CGI escaped URI. 

So the configuration command will look something like:

```
bundle config mirror.https://private.gem.host http://localhost:9292/upstream/$(ruby -rcgi -e 'puts CGI.escape("https://user:password@private.gem.host")')
```

jebentier:
Recently tried doing the above to mirror our Gemfury repo which uses basic auth as well but just an api token of the form `https://<api_token>@gem.fury.io/<repo>`.  Because the basic auth isn't of the form `user:password`, Faraday by default wasn't respecting it in the URL. I'm putting together a small PR that adds support for this, which should help anyone that using a private repo service like gemfury.

PR #317, opened 2022-06-08
Respect auth set in upstream host uri

jebentier:
# Description:

Some private gem repos like [Gemfury](https://gemfury.io) use Basic Auth with just an api key rather than a username and password (ex. `https://<api_key>@gem.fury.io/<repo>`). Setting up upstream mirroring for this case should work as described in #247, but because `Faraday` does not respect Basic Authentication being passed in the URL unless its of the `user:pass` format, some additional logic was needed to support this use case.

______________

# Tasks:

-  Write tests

### DIFF
--- a/lib/gemstash/http_client.rb
+++ b/lib/gemstash/http_client.rb
@@ -35,11 +35,16 @@ module Gemstash
         config.adapter :net_http
         config.options.timeout = gemstash_env.config[:fetch_timeout]
       end
+
+      client.basic_auth(upstream.user, upstream.password) if upstream.auth?
+
       user_agent = "#{upstream.user_agent} " unless upstream.user_agent.to_s.empty?
       user_agent = user_agent.to_s + DEFAULT_USER_AGENT
 
       new(client, user_agent: user_agent)
     end
+
+    attr_reader :client
 
     def initialize(client = nil, user_agent: nil)
       @client = client
--- a/lib/gemstash/upstream.rb
+++ b/lib/gemstash/upstream.rb
@@ -32,7 +32,7 @@ module Gemstash
     end
 
     def auth?
-      !user.to_s.empty? && !password.to_s.empty?
+      !user.to_s.empty? || !password.to_s.empty?
     end
 
     # Utilized as the parent directory for cached gems
--- a/spec/gemstash/http_client_spec.rb
+++ b/spec/gemstash/http_client_spec.rb
@@ -22,6 +22,26 @@ RSpec.describe Gemstash::HTTPClient do
     @other_server.stop
   end
 
+  describe '#for' do
+    context 'when user:pass auth is in the upstream url' do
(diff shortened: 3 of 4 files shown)


lib/gemstash/http_client.rb
```diff
@@ -35,12 +35,17 @@ def self.for(upstream)
         config.adapter :net_http
         config.options.timeout = gemstash_env.config[:fetch_timeout]
       end
+
+      client.basic_auth(upstream.user, upstream.password) if upstream.auth?
```
olleolleolle, line 39: Minor: perhaps we can check if the basic_autj method is defined, before using it. There's been a new version, which has another variant method. Faraday  has a nite about that in its website.

olleolleolle: A note.

olleolleolle:
Awesome addition!

lib/gemstash/http_client.rb
jebentier: I can definitely add that if it's warranted. My thinking was this kind of check was unnecessary since the gemspec is locking the Faraday gem to `~> 0.9` and the new interface only exist in `>= 1.0`, and unit test coverage will show that the major version bump of Faraday will break existing functionality and force the necessary code change.

olleolleolle: Cool, cool, we'll get the relevant deprecation warnings when we get there.

olleolleolle:
<details>

```
spec/gemstash/http_client_spec.rb:25:12: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
  describe '#for' do
           ^^^^^^
spec/gemstash/http_client_spec.rb:26:13: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
    context 'when user:pass auth is in the upstream url' do
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/gemstash/http_client_spec.rb:32:13: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
    context 'when api_key auth is in the upstream url' do
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/gemstash/http_client_spec.rb:38:13: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
    context 'when no auth is included in the upstream url' do
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

</details>

olleolleolle:
I noted that the JRuby 9.2 build had failures from having a too-old RubyGems preinstalled: now we updated, and we went from "erroring the build" to "one test failure".

PR #326, opened 2023-01-23
Change the response code for "/versions" from 403 to 404

hendric-dev:
This allows to use Renovate with Gemstash.
Renovate changed their update logic to use the `/versions` endpoint, breaking the compatibility with Gemstash.
However the old logic is still used as a fallback in case the endpoint returns a 404.

Related PR: https://github.com/renovatebot/renovate/pull/16312

### DIFF
--- a/lib/gemstash/gem_source/private_source.rb
+++ b/lib/gemstash/gem_source/private_source.rb
@@ -47,7 +47,7 @@ module Gemstash
       end
 
       def serve_versions
-        halt 403, "Not yet supported"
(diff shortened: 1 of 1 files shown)


jume-dev:
@indirect Will there be a release with this fix soonish?

indirect:
@jume-dev there was a release. 👍🏻

johnjelinek:
Good timing, because the Dependency API is going away: https://blog.rubygems.org/2023/04/07/dependency-api-deprecation-delayed.html

PR #328, opened 2023-01-27
Update Ruby versions

### LINKED ISSUE #325: Use JRuby 9.4 in CI
https://www.jruby.org/2022/11/23/jruby-9-4-0-0.html is released. Perhaps we should have that in the CI?

indirect:
Adds Ruby 3.2 to CI, drops Ruby 2.4, 2.5, and 2.6 as they have been EOLed.

Closes #325.


olleolleolle (approved):
I read the changeset - and it looks good, builds green, and drops unused versions like it says. 

More expressive in places. 

Thanks for the care put into this. It means a lot.

olleolleolle:
In light of the final commit, ought we to re-ignore Gemfile.lock?

indirect:
Yes! Now that the tests pass, cleaning up seems good. 😅

PR #333, opened 2023-03-03
Add .rake_github_token to .gitignore

olleolleolle:
# Description:

This PR git-ignores a file which can contain credentials.

The token file is used when releasing, to regenerate the CHANGELOG.md file, using a rake task "changelog".

Unrelated to this change: When I ran that task today, it failed, with a rate-limiting error, and I was unable to update the file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 /pkg/
 /spec/reports/
 /tmp/
+.rake_github_token


olleolleolle:
@indirect I now found https://github.com/rubygems/gemstash/commit/b897639fbc76f41eac696ba6cd567cd80e057527 which mentioned the "can not auth, even with token".

PR #335, opened 2023-03-22
Updates for Puma 6

MSP-Greg:
# Description:

Current gemspec is using `"puma", "~> 4.0"`.  Puma 4 is incompatible with OpenSSL 3.  This updates to `"puma", "~> 6.1"`.

I believe I made all the needed changes, as the Puma 6 API is somewhat different that Puma 4.


olleolleolle (approved):
Well-used context-preserving comments about Puma versions.

lib/gemstash/logging.rb
```diff
@@ -1,7 +1,12 @@
 # frozen_string_literal: true
 
 require "logger"
-require "puma/events"
+
+begin
+  require "puma/log_writer" # Puma 6
+rescue LoadError
+  require "puma/events"
```
olleolleolle, line 8: I wonder, when you put in ~> 6.1 in gemspec, will anyone ever run an older Puma with this code?

indirect:
Thanks!

Issue #336, opened 2023-04-11
Retirement of Dependencies API breaks clients that cannot access rubygems


tedgarb:
Discovered this today, believed to be fallout from https://blog.rubygems.org/2023/02/22/dependency-api-deprecation.html . When I try to install a gem using my mirror (e.g. net-ldap), the following is the verbose logs of installing on a client that does not have access to rubygems.org:

```gem install --no-document --source http://my-mirror.com/rubygems/ net-ldap -V
HEAD http://my-mirror.com/rubygems/api/v1/dependencies
200 OK
GET http://my-mirror.com/rubygems/api/v1/dependencies?gems=net-ldap
404 Not Found
GET http://my-mirror.com/rubygems/prerelease_specs.4.8.gz
302 Found
GET http://my-mirror.com/rubygems/specs.4.8.gz
302 Found
ERROR:  Could not find a valid gem 'net-ldap' (>= 0), here is why:
          Unable to download data from https://rubygems.org/ - Errno::ENETUNREACH: Failed to open TCP connection to rubygems.org:443 (Network is unreachable - connect(2) for "rubygems.org" port 443) (https://rubygems.org/specs.4.8.gz)
          Unable to download data from http://my-mirror.com/rubygems/ - Errno::ENETUNREACH: Failed to open TCP connection to rubygems.org:443 (Network is unreachable - connect(2) for "rubygems.org" port 443) (http://my-mirror.com/rubygems/specs.4.8.gz)
GET http://my-mirror.com/rubygems/latest_specs.4.8.gz
302 Found
GET http://my-mirror.com/rubygems/prerelease_specs.4.8.gz
302 Found
```

with this in the gemstash server log:
```[2023-04-11 11:28:38 -0400] - INFO - Fetching dependencies: net-ldap
[2023-04-11 11:28:38 -0400] - ERROR - 2023-04-11 11:28:38 - Gemstash::WebError - The dependency API is going away. See https://blog.rubygems.org/2023/02/22/dependency-api-deprecation.html for more information:
	/usr/local/share/gems/gems/gemstash-2.2.2/lib/gemstash/http_client.rb:62:in `get'
	/usr/local/share/gems/gems/gemstash-2.2.2/lib/gemstash/dependencies.rb:80:in `fetch_from_web'
	/usr/local/share/gems/gems/gemstash-2.2.2/lib/gemstash/dependencies.rb:43:in `fetch'
[...]

PR #337, opened 2023-06-15
Make UpstreamSource intermediate specs from upstream

tedgarb:
As discovered in #336, the retirement of the v1 dependencies API revealed that UpstreamSource, when the dependencies API is unavailable, tells the client to go talk to rubygems directly for the spec files. This model does not work in general, because there is no guarantee that the client can access rubygems directly. This commit changes UpstreamSource to intermediate the connection

# Description:

______________

# Tasks:

-  Write tests

### DIFF
--- a/lib/gemstash/gem_source/upstream_source.rb
+++ b/lib/gemstash/gem_source/upstream_source.rb
@@ -128,6 +128,21 @@ module Gemstash
         serve_cached(id, :gem)
       end
 
+      def serve_latest_specs
+        http_client = http_client_for(upstream)
+        http_client.get("latest_specs.4.8.gz")
+      end
+
+      def serve_prerelease_specs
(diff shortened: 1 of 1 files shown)


indirect:
Have you tested this? I don't think it will allow installation of gems because the `/quick/Marshal.4.8/` specs aren't also proxied.

tedgarb:
quick/Marshal is handled in that class via the cache mechanism in https://github.com/rubygems/gemstash/blob/523a7e60a7a3b9184fc3ab1d062a5c8534bc9ac8/lib/gemstash/gem_source/upstream_source.rb#L123 

This patch is identical to the code I am running in production in my environment today, which is letting me install gems and to search for them

indirect:
Ah, got it. Thanks for explaining!

PR #339, opened 2023-07-24
Add support for upstream auth with ENV variables

CiTroNaK:
# Description:

Allowing to set auth credentials for upstreams on the Gemstash instance with `ENV` variables (`GEMSTASH_<HOST>`), that can contain basic auth or API key.

This is mainly POC, and I will happily update the PR based on your feedback and add this to the docs.

## Use case

The primary use case is to have only one place where the credentials for private/licensed sources are set without storing them inside the Gemfile or distributing them to all developers. That should significantly reduce the time needed to rotate or set (distribute) new credentials (for each developer in the team or CI and other services that need to access it). This also should help to prevent leaking the credentials.

Gemstash instance should be secured using (at least) basic authentication. With this setup, anyone could point their Gemfile to it and be able to fetch private/licensed gems. Developers would only set credentials to the Gemstash instance locally.

## Example

Version 2.2.2 with this commit was used for testing.

### Setup gemstash
```bash
GEMSTASH_GEMS__CONTRIBSYS__COM=user:pass gemstash start --no-daemonize --config-file config.yml.erb
```

### Gemfile
With a gem under auth (Sidekiq Pro) without setting the credentials in the upstream.

```ruby
source "http://localhost:9292/"

ruby "3.2.2"

source "http://localhost:9292/upstream/gems.contribsys.com" do
  gem "sidekiq-pro", "~> 5.5.7"
end
```

### Bundle

```bash
➜ bundle install           
Fetching gem metadata from http://localhost:9292/upstream/gems.contribsys.com..
Fetching gem metadata from http://localhost:9292/............
Using bundler 2.4.10
Using connection_pool 2.4.1
Using rack 2.2.7
Using redis 4.8.1
Using sidekiq 6.5.9
Fetching sidekiq-pro 5.5.8
Installing sidekiq-pro 5.5.8
Bundle complete! 1 Gemfile dependency, 6 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
```

## Next steps (if this PR will be accepted)

- [x] Update changelog
- [x] Update docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Add support for upstream auth with ENV variables ([#339](https://github.com/rubygems/gemstash/pull/339), [@CiTroNaK](https://github.com/CiTroNaK))
+
 ## 2.4.0 (2023-09-27)
 
 ### Changes
--- a/docs/gemstash-multiple-sources.7.md
+++ b/docs/gemstash-multiple-sources.7.md
@@ -49,6 +49,10 @@ gem "rubywarrior"
 source "http://localhost:9292/upstream/#{CGI.escape("https://my.gem-source.local")}" do
   gem "my-gem"
 end
+
+source "http://localhost:9292/upstream/my-other.gem-source.local" do
+  gem "my-other-gem"
+end
 ```
 
 Notice the `CGI.escape` call in the second source. This is important, as
@@ -56,6 +60,54 @@ it properly URL escapes the source URL so Gemstash knows what gem source
 you want. The `/upstream` prefix tells Gemstash to use a gem source
 other than the default source. You can now bundle with the additional
 source.
+
+Notice that the third source doesn't need to be escaped.
+This is because the `https://` is used by default when no scheme is set,
+and the source URL does not contain any chars that need to be escaped.
+
+## Authentication with Multiple Sources
+
+You can use basic authentication or API keys on sources directly in Gemfile
+or using ENV variables on the Gemstash instance.
+
+Example `Gemfile`:
+```ruby
+# ./Gemfile
+require "cgi"
+source "http://localhost:9292"
+
+source "http://localhost:9292/upstream/#{CGI.escape("user:password@my.gem-source.local")}" do
+  gem "my-gem"
+end
+
+source "http://localhost:9292/upstream/#{CGI.escape("api_key@my-other.gem-source.local")}" do
+  gem "my-other-gem"
+end
+```
+
+If you set `GEMSTASH_<HOST>` ENV variable with your authentication information,
+you can omit it from the `Gemfile`:
+
+```ruby
+# ./Gemfile
+source "http://localhost:9292"
+
+source "http://localhost:9292/upstream/my.gem-source.local" do
+  gem "my-gem"
+end
+```
+
+And run the Gemstash with the credentials set in an ENV variable:
+
+```bash
+GEMSTASH_MY__GEM___SOURCE__LOCAL=user:password gemstash start --no-daemonize --config-file config.yml.erb
+```
+
+The name of the ENV variable is the uppercase version of the host name,
+with all `.` characters replaced with `__`, all `-` with `___` and a `GEMSTASH_` prefix 
+(it uses the same syntax as [Bundler](https://bundler.io/v2.4/man/bundle-config.1.html#CREDENTIALS-FOR-GEM-SOURCES)).
+
+Example: `my.gem-source.local` => `GEMSTASH_MY__GEM___SOURCE__LOCAL`
 
 ## Redirecting
 
--- a/lib/gemstash/upstream.rb
+++ b/lib/gemstash/upstream.rb
@@ -11,10 +11,12 @@ module Gemstash
 
     attr_reader :user_agent, :uri
 
-    def_delegators :@uri, :scheme, :host, :user, :password, :to_s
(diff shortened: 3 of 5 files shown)


indirect:
I really like this approach to solve the gem licensing problem! Something like [Tailscale](https://tailscale.com) could also offer an alternative to basic auth or other Gemstash credentials. I wonder if there's a way for us to make this less painful for RubyGems users generally... thanks for the inspiration, and I'll keep thinking about this general issue even after we finish and merge this change.

Would you feel ok about defaulting to https for upstream sources? So the URL could then be `http://localhost:9292/upstream/gems.contribsys.com`. Overall, I am happy with this and would be glad to merge it.

CiTroNaK:
Hi @indirect, I am glad to read it 🙂 I've updated the PR with the default scheme and updated the changelog and docs.

Please, let me know if you find something that should be updated.

Thank you!

hersha:
Just my 2 cents but why not just use the bundler envvar naming convention? Then I can just copy things wholesale from my local env and not have to worry about remembering to prefix it with gemstash.

CiTroNaK:
@hersha Hi, it uses the same ENV naming conventions as Bundler, so you only need to replace the `BUNDLE_` prefix with a `GEMSTASH_`.

We could add support for both, but it wouldn't be beneficial. Having correctly named ENV variables helps with their management. And who knows, maybe there is a case where having them set for Bundler on the same machine would not be desirable.

docs/gemstash-multiple-sources.7.md
```diff
@@ -57,6 +61,53 @@ you want. The `/upstream` prefix tells Gemstash to use a gem source
...
+```
+
+And run the Gemstash with the credentials set in an ENV variable:
+
+```bash
+GEMSTASH_MY__GEM___SOURCE__LOCAL=user:password gemstash start --no-daemonize --config-file config.yml.erb
+```
+
+The name of the ENV variable is the uppercase version of the host name,
+with all `.` characters replaced with `__`, all `-` with `___` and a `GEMSTASH_` prefix.
```
olleolleolle, line 107: Minor: Perhaps we can link to Bundler's documentation about their host-in-environment variables? https://bundler.io/v2.4/man/bundle-config.1.html#CREDENTIALS-FOR-GEM-SOURCES

lib/gemstash/upstream.rb
```diff
@@ -40,12 +41,41 @@ def host_id
...
   private
 
     def hash
       Digest::MD5.hexdigest(to_s)
     end
 
+    def env_auth_user
+      return unless env_auth
+
+      env_auth.split(":").first
```
olleolleolle, line 61:
```suggestion
      env_auth.split(":", 2).first
```

lib/gemstash/upstream.rb
```diff
@@ -11,10 +11,11 @@ class Upstream
 
     attr_reader :user_agent, :uri
 
-    def_delegators :@uri, :scheme, :host, :user, :password, :to_s
+    def_delegators :@uri, :scheme, :host, :to_s
 
     def initialize(upstream, user_agent: nil)
-      @uri = URI(CGI.unescape(upstream.to_s))
+      url = upstream.to_s.start_with?("http") ? upstream.to_s : "https://#{upstream}"
```
olleolleolle, line 17: Q: Would it be prudent to do `.start_with("http://")` or `.start_with("https://")` so that inputs that are bare domain names that begin with http, such as `httpstatus.es` etc, are covered?

olleolleolle (commented):
Thanks for supporting Gemstash with this quality of life improver.

CiTroNaK:
Hi @olleolleolle, thank you for the review. I rebased and updated the PR.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Add support for upstream auth with ENV variables ([#339](https://github.com/rubygems/gemstash/pull/339), [@CiTroNaK][])
```
olleolleolle, line 3: Minor thing: The Rake task `changelog` has a `.citrus` grammar which looks for _full_ username links, rather than "reference-style links".

```suggestion
- Add support for upstream auth with ENV variables ([#339](https://github.com/rubygems/gemstash/pull/339), [@CiTroNaK](https://github.com/CiTroNaK))
```

spec/gemstash/upstream_spec.rb
```diff
@@ -55,19 +65,43 @@
...
-  end
-
   it "has a nil user agent if not provided" do
     expect(Gemstash::Upstream.new("https://rubygems.org/").user_agent).to be_nil
   end
 
   it "supports getting user agent" do
     expect(Gemstash::Upstream.new("https://rubygems.org/",
-                                  user_agent: "my_user_agent").user_agent).to eq("my_user_agent")
+      user_agent: "my_user_agent").user_agent).to eq("my_user_agent")
```
olleolleolle, line 74: RuboCop complains about this line, perhaps you can autofix it, perhaps you can return it to what it was before?

CiTroNaK: Sorry, an automatic "fix" from my editor... 🤦‍♂️ It is fixed now.

olleolleolle (approved):
Let's get this merged!

Thank you for being so patient and generous with your time, @CiTroNaK!

:hearts: :yellow_heart: :green_heart: :blue_heart: :purple_heart:

olleolleolle:
@CiTroNaK Now released as v2.5.0. Cheers!

PR #341, opened 2023-08-29
CI: Rename dependabot.yml to not be a "workflow"

olleolleolle:
# Description:

This change places the Dependabot configuration file in the right directory.

This file is not a workflow, so it should move out of the GitHub Actions workflows directory, and into .github/.


olleolleolle:
That worked! And there were immediate Dependabot PRs.

PR #365, opened 2023-09-25
authorize: support naming and listing authorizations

### LINKED ISSUE #364: Authorization tokens are easy to lose track of
We've been using the private gem functionality of gemstash for a number of years now, with great success. However, as both our team and our CI systems have changed a little over time, we're hitting a pain point: we need to keep careful notes of which token belongs to each developer or system, so we can revoke tokens when they're no longer used. This feels like a feature that could be easily added to gemstash, and if we agree on a design I'd be happy to add it. I'm thinking something really simple:

- The ability to name or label a given authorization.
- The ability to list authorizations, with their names and capabilities

Thoughts?

kyrofa:
Over time, gemstash authorizations turn into a pile of opaque tokens and associated permissions. Keeping track of which token belongs to which developer or automated system is required to enable proper revocation or rotation. Today, that requires the use of an external system, which has the added problem of duplicating the key itself.

This PR resolves #364, adding support for keeping this association within Gemstash itself by allowing for authorizations to be named. It also adds support for listing authorizations, so it's easy to determine which key needs to be revoked, rotated, or otherwise updated. See the manpage updates for details on use.

### DIFF
--- a/lib/gemstash/migrations/05_authorization_names.rb
+++ b/lib/gemstash/migrations/05_authorization_names.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table :authorizations do
+      add_column :name, String, :size => 191
+      add_index [:name], :unique => true
+    end
+  end
+end
--- a/spec/gemstash/cli/authorize_spec.rb
+++ b/spec/gemstash/cli/authorize_spec.rb
@@ -239,7 +239,7 @@ RSpec.describe Gemstash::CLI::Authorize do
       expect { Gemstash::CLI::Authorize.new(cli).run }.to raise_error(Gemstash::CLI::Error)
     end
 
-    it "handles missing authorization" do
+    it "raises error when authorization is missing" do
       Gemstash::Authorization.authorize("auth-key-all", "all", "auth all")
       Gemstash::Authorization.authorize("auth-key-push", %w[push], "auth push")
       cli_options[:name] = "missing name"
--- a/spec/gemstash/cli/authorize_spec.rb
+++ b/spec/gemstash/cli/authorize_spec.rb
@@ -58,6 +58,19 @@ RSpec.describe Gemstash::CLI::Authorize do
     end
   end
 
+  context "authorizing with a name without specifying the key" do
+    let(:cli_options) { { :name => "test auth" } }
+
+    it "saves a new key with the given name" do
+      expect(SecureRandom).to receive(:hex).and_return("new-auth-key")
+      Gemstash::CLI::Authorize.new(cli).run
+      expect(@said).to include("new-auth-key")
+      authorization = Gemstash::Authorization["new-auth-key"]
+      expect(authorization.all?).to be_truthy
+      expect(authorization.name).to eq "test auth"
+    end
+  end
+
   context "authorizing without specifying the key and with permissions" do
     it "outputs the new key and authorizes for the given permissions" do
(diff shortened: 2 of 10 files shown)


lib/gemstash/migrations/05_authorization_names.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table :authorizations do
+      add_column :name, String, :size => 191
```
olleolleolle, line 6: Size seems oddly specific, am I missing something?

spec/gemstash/cli/authorize_spec.rb
```diff
@@ -116,6 +136,18 @@
       Gemstash::CLI::Authorize.new(cli).run
       expect(Gemstash::Authorization["auth-key"]).to be_nil
     end
+
+    it "combined with --list results in an error" do
+      cli_options[:list] = true
+      expect { Gemstash::CLI::Authorize.new(cli).run }.to raise_error(Gemstash::CLI::Error)
+      expect(Gemstash::Authorization["auth-key"]).to be # Auth was not actually removed
```
olleolleolle, line 143: TIL about this very bare form of Matcher.

spec/gemstash/cli/authorize_spec.rb
```diff
@@ -148,4 +180,71 @@
...
+    it "errors when auth is specified by both name and key" do
+      Gemstash::Authorization.authorize("auth-key-all", "all", "auth all")
+      Gemstash::Authorization.authorize("auth-key-push", %w[push], "auth push")
+      cli_options[:name] = "auth push"
+      cli_options[:key] = "auth-key-push"
+
+      expect { Gemstash::CLI::Authorize.new(cli).run }.to raise_error(Gemstash::CLI::Error)
+    end
+
+    it "handles missing authorization" do
```
olleolleolle, line 242: Perhaps this should express "raises an error on missing" so that we describe its behaviour.

olleolleolle (commented):
Wonderful, thank you! A thorough feature, well-described. @kyrofa, this is good.

Edit: if you please, add a CHANGELOG description, should you have one.

lib/gemstash/migrations/05_authorization_names.rb
kyrofa: Ha! We share that thought. I was honestly just duplicating the [existing pattern for String columns](https://github.com/rubygems/gemstash/blob/main/lib/gemstash/migrations/02_authorizations.rb#L7-L8) in the authorizations table, because they're magic numbers there as well. This number means nothing to me. Would you prefer something different?

spec/gemstash/cli/authorize_spec.rb
kyrofa: Same. I stole it from another test! One of the beautiful things about open source :heart: .

spec/gemstash/cli/authorize_spec.rb
kyrofa: Sure thing, fixed.

kyrofa:
> Edit: if you please, add a CHANGELOG description, should you have one.

Done, does that look right?

olleolleolle (approved):
Grand, let's get this one out!

lib/gemstash/migrations/05_authorization_names.rb
olleolleolle: I learned, by inspecting the git blame navigation: https://github.com/rubygems/gemstash/pull/64

PR #373, opened 2023-09-29
configuration: support GEMSTASH_CONFIG environment variable

### LINKED ISSUE #369: Support config file in /etc
We run gemstash in a read-only Docker container. A persistent volume is mounted to a particular directory, and that's the base path for Gemstash. We ship the gemstash config in `/etc/gemstash/config.yml`, because it's a system-wide config and that's generally where those live. It's a bit annoying to constantly have to include `--config-file` when adding new authorizations and whatnot, though. How would folks feel about checking `/etc/gemstash/config.yml` in addition to the `~/.gemstash/config.yml` that is checked today? Or, barring that, having gemstash support an environment variable to point to its config file?

kyrofa:
Some gemstash configurations don't have a `/home/gemstash` directory (e.g. minimal Docker installations), and prefer to store their Gemstash configuration elsewhere (e.g. in a system-wide location such as `/etc/gemstash/config.yml`). Today, such use-cases require every Gemstash command to be accompanied by the `--config-file` CLI option, and forgetting to do so causes Gemstash to use its default settings.

This PR resolves #369, smoothing the wrinkles for this use-case, by supporting defining the path to one's configuration in the `GEMSTASH_CONFIG` environment variable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Changes
+
+  - Support `GEMSTASH_CONFIG` environment variable ([#369](https://github.com/rubygems/gemstash/pull/369), [@kyrofa](https://github.com/kyrofa))
+
 ## 2.5.0 (2023-09-28)
 
 - Add support for upstream auth with ENV variables ([#339](https://github.com/rubygems/gemstash/pull/339), [@CiTroNaK](https://github.com/CiTroNaK))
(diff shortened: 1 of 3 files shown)


indirect (approved):

olleolleolle:
Yes! Thank you!

:hearts: :yellow_heart: :green_heart: :blue_heart: :purple_heart:

PR #376, opened 2023-10-07
CLI daemonized cannot be set to true. fixes #359

### LINKED ISSUE #359: Drop --daemonize which is not supported by stock Puma
Perhaps it is time for this project not to default to "daemonize on". EDIT: time to remove this command-line option which is unsupported.

Puma 5 stopped daemonize by default.

Context:
https://github.com/kigster/puma-daemon#why-was-it-removed-from-puma-v5

thiagopintodev:
I needed to run an gemstash when I ran into this problem.
I hope this fix is well received.

### DIFF
--- a/lib/gemstash/cli.rb
+++ b/lib/gemstash/cli.rb
@@ -75,7 +75,7 @@ module Gemstash
     end
 
     desc "start", "Starts your gemstash server"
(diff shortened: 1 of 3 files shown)


olleolleolle:
Thanks for using and supporting Gemstash!

I think we should just remove this footgun option.

thiagopintodev:
closing in favor of #378

Issue #380, opened 2023-10-11
Release a 2.7.1 supporting a --no-daemonize option that only outputs a warning


olleolleolle:
>> Ah, then I am mistaken indeed, and I understand where you're coming from now. Thank you for your patience!
>
> Thanks for you patience as well, let's release the `2.7.1` soon adding back no-op `--no-daemonize` with warning.

_Originally posted by @simi in https://github.com/rubygems/gemstash/issues/378#issuecomment-1755974160_

simi:
fixed by #381

PR #378, opened 2023-10-07
Remove the daemonize feature

olleolleolle:
# Description:

This is a Pull Request which removes the `--daemonize` option, its documentation, and all related mentions.

Puma no longer supports it. There are new, other ways to do that.

This closes #359 with a more-final change.

Perhaps this means that there we may need new ways of enabling a Puma::CommonLogger, of enabling non-stdout output (that is "setup_logger") in the Gemstash server.

Now, gemstash operates "in front, in the terminal it is started".

I will abide by the [code of conduct](https://github.com/gemstash/gemstash/blob/master/CODE_OF_CONDUCT.md).

PR #381, opened 2023-10-12
Reinstate nonfunctional --no-daemonize, --daemonize options that warn

olleolleolle:
# Description:

Reinstate a `--daemonize` and `--no-daemonize` which only warn.

Follow-up to #378 and fixes the problem in #380.

# Tasks:

-  Write tests

### DIFF
--- a/lib/gemstash/cli.rb
+++ b/lib/gemstash/cli.rb
@@ -75,6 +75,8 @@ module Gemstash
     end
 
     desc "start", "Starts your gemstash server"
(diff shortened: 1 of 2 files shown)


simi (approved):

olleolleolle:
See https://github.com/ruby/psych/issues/655 for the reason why JRuby fails its build.

olleolleolle:
As a follow-up, I have now release gemstash-2.7.1-java.gem, too

PR #392, opened 2024-07-09
Add compact index support for private sources

segiddins:
# Description:

Ensured working via https://github.com/rubygems/gem_server_conformance

- [x] Blocked by https://github.com/jruby/activerecord-jdbc-adapter/pull/1149
- [x] 2 postgres integration specs failing on CI only

______________

# Tasks:

-  Write tests


indirect (approved):
looking good!

I'm also interested to hear if this gave you any ideas about how we could improve the compact index gem for consumers, so their implementations could potentially be shorter/easier.

lib/gemstash/compact_index_builder.rb
```diff
@@ -0,0 +1,257 @@
...
+
+      def build_result
+        names = DB::Rubygem.db[<<~SQL.squish].map {|row| row[:name] }
+          SELECT name
+          FROM rubygems
+          LEFT JOIN versions ON versions.rubygem_id = rubygems.id
+          WHERE versions.indexed = true
+          GROUP BY name
+          HAVING COUNT(versions.id) > 0
+          ORDER BY name
```
olleolleolle, line 245: I wonder, can this become an inner join rather than a LEFT JOIN? That is, excluding rubygems rows that have no related row in versions. Then, we'd not need the `HAVING COUNT(versions.id) > 0`.

Am I missing something here?

technicalpickles:
Is there any extra configuration needed to enable this? Wanted to test locally, but see this when trying to bundle install against it:

```
HTTP GET http://localhost:9292/private/versions
HTTP 404 Not Found http://localhost:9292/private/versions
Bundler::Fetcher::FallbackError: Gem::Net::HTTPNotFound: http://localhost:9292/private/versions
```

segiddins:
@technicalpickles shouldn't be, what's showing up in the server logs?

technicalpickles:
Startup:
```
❯ bin/gemstash start                                                                                                                                                                      ─╯
Starting gemstash!
[2024-07-31 17:10:38 -0400] - INFO - [19490] Puma starting in cluster mode...
[2024-07-31 17:10:38 -0400] - INFO - [19490] * Puma version: 6.4.2 (ruby 3.3.2-p78) ("The Eagle of Durango")
[2024-07-31 17:10:38 -0400] - INFO - [19490] *  Min threads: 0
[2024-07-31 17:10:38 -0400] - INFO - [19490] *  Max threads: 16
[2024-07-31 17:10:38 -0400] - INFO - [19490] *  Environment: development
[2024-07-31 17:10:38 -0400] - INFO - [19490] *   Master PID: 19490
[2024-07-31 17:10:38 -0400] - INFO - [19490] *      Workers: 1
[2024-07-31 17:10:38 -0400] - INFO - [19490] *     Restarts: (✔) hot (✔) phased
[2024-07-31 17:10:38 -0400] - INFO - [19490] * Listening on http://0.0.0.0:9292
[2024-07-31 17:10:38 -0400] - INFO - [19490] Use Ctrl-C to stop
[2024-07-31 17:10:38 -0400] - INFO - [19490] ! WARNING: Detected running cluster mode with 1 worker.
[2024-07-31 17:10:38 -0400] - INFO - [19490] ! Running Puma in cluster mode with a single worker is often a misconfiguration.
[2024-07-31 17:10:38 -0400] - INFO - [19490] ! Consider running Puma in single-mode (workers = 0) in order to reduce memory overhead.
[2024-07-31 17:10:38 -0400] - INFO - [19490] ! Set the `silence_single_worker_warning` option to silence this warning message.
/Users/josh.nichols/workspace/gemstash/lib/gemstash/web.rb:10: warning: Skipping set of ruby2_keywords flag for initialize (method accepts keywords or method does not accept argument splat)
[2024-07-31 17:10:38 -0400] - INFO - [19490] - Worker 0 (PID: 19492) booted in 0.36s, phase: 0
```

During `bundle update bigdecimal --conservative`:

```
[2024-07-31 17:10:47 -0400] - INFO - Rewriting '/private/versions' to '/versions'
[2024-07-31 17:10:47 -0400] - INFO - Rewriting '/private/api/v1/dependencies' to '/api/v1/dependencies'
[2024-07-31 17:10:47 -0400] - INFO - Rewriting '/private/api/v1/dependencies?gems=redacted-gem' to '/api/v1/dependencies?gems=redacted-gem'
[2024-07-31 17:10:47 -0400] - INFO - Querying dependencies: redacted-gem
[2024-07-31 17:10:47 -0400] - INFO - Rewriting '/private/api/v1/dependencies?gems=bigdecimal' to '/api/v1/dependencies?gems=bigdecimal'
[2024-07-31 17:10:47 -0400] - INFO - Querying dependencies: bigdecimal
```

segiddins:
I'll revisit this once there's a version of sqlite for jruby that supports what I used here

martinemde:
Hey @segiddins, the jdbc-sqlite3 seems like a long term blocker at this point. Is the performance greatly affected by not having string_agg? I'd be willing to fix it up to see if we can get this in if you don't have time, but I could use some context about that choice.

segiddins:
It's the only way to get an array out of SQLite unfortunately. I wish we knew what jruby usage was like :/

technicalpickles:
Would it feasible to disable compact index on jruby until the blocker is fixed? That way, non-jruby users would get the benefit. We could also keep a PR open to enable it on jruby, so we have a reminder and/or place to validate fixes.

simi:
CI logs are not reachable anymore. Would you mind to rebase this and update deps? Seems there are some news at https://github.com/jruby/activerecord-jdbc-adapter/pull/1149#issuecomment-2389483680.

martinemde:
When I was looking closely at Ruby compact index results, they were coming out ordered differently even with identical code (and even aggregated as a separate query with its own sort) so I suspect jdbc-sqlite3 is not behaving identically to cruby's adapters. The order might be because of time field precision differences since we sort by created_at. Just a hunch.

lib/gemstash/migrations/06_compact_index.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table :versions do # TODO: backfill info_checksum, sha256, required_ruby_version, required_rubygems_version
```
segiddins, line 5: @indirect any thoughts on how best to do this, now that the PR is passing?

schema.rb
segiddins, line 1: - [ ] delete

lib/gemstash/migrations/06_compact_index.rb
martinemde: Should we add a `gemstash update` or `gemstash sync` command that would be a standard idempotent way to reindex/update after a new version is released?

Have we done a migration like this before?

indirect: yeah, I think a command to `migrate` or `update` is the way to go, with some docs about how to update to the newer version.

rsanheim:
> What do you think of skipping conformance tests for jruby and releasing compact index as experimental on jruby for now?

This sounds like a great plan, as I know compact index support would help a lot for our usage of gemstash at Doximity (and we are not on jruby). 

Also, getting this merged would possibly nudge jruby/genstash users to debug and fix any follow up issues if needed. 😏

segiddins:
It's passing on jruby now ;)

martinemde (approved):
Only small style stuff and calling out boilerplate comments. Looks great.

lib/gemstash/migrations/06_compact_index.rb
technicalpickles: Are we sure we need an explicit command? I found `Gemstash::Env.migrate` which handles running migrations. That is called in `Gemstash::Env.db`, so basically anytime the db is accessed it is automatically migrated before rturning a connection.

A1bi:
We needed our company-internal gemstash instance to support the compact index API, mainly for gem checksums. We therefore decided to apply the changes from this branch to our instance. This has been running flawlessly in production use for a few months now. We only needed to write a small script to backfill checksums for existing gems since #412 is still missing that part.

Any way we can help getting this over the finish line for a proper release?

technicalpickles:
We've internally migrated away from gemstash, so I haven't been working on this.

Issue #396, opened 2024-08-03
`bin/gemstash` fails with `Could not find command "readme"`


technicalpickles:
This is specifically about running gemstash in development, but I'm able to do things like `bin/gemstash start`, but `bin/gemstash` or anything that would do help usage fails like:

```
❯ bin/gemstash                                                                                                                                                                                                  
Could not find command "readme".
```

This is the only place I can find readme mentioned:

https://github.com/rubygems/gemstash/blob/e0e1b4b920ac29b9f777e0cd0b9325e3d73f5259/lib/gemstash/cli.rb#L40-L49

olleolleolle:
Perhaps the casing changed, and we need to support the uppercase README (md) file, here?

technicalpickles:
Forgot to post this, but I was able to confirm that running `rake doc` does create the files it's looking for, ie `/Users/josh.nichols/workspace/gemstash/lib/gemstash/man/gemstash-readme.7`

This is definitely specific to local development. I'm not sure what, if anything, should happen to read from the local files in this case 🤔

olleolleolle:
@technicalpickles Yeah, but it _is_ a weird location to use, isn't it? `lib/gemstash/man/` rather than `man/`, I mean.

olleolleolle:
@technicalpickles Thanks, coming back to this, perhaps the message could explain "if you're in local dev, please regenerate docs"?

technicalpickles:
Yeah, that makes sense to me 👍🏻  I was trying to think how you might determine that, and I think checking `$0 == "bin/gemstash"` might be the easiest way?

PR #397, opened 2024-08-05
Doc task: land man page into man/

olleolleolle:
# Description:

We use Pandoc and groff to make man page files in roff and text formats.

Those files were generated into the lib/gemstash/man directory, where the CLI tools aren't looking for them.

#396 shows what that leads to: no man pages are found.

This PR changes the output directory of the export to be `man/`.

---

### DIFF
--- a/rakelib/doc.rake
+++ b/rakelib/doc.rake
@@ -71,7 +71,7 @@ class Doc
     end
 
     def export_to_man_and_txt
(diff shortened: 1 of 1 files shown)


olleolleolle:
Of course, there were subtleties, and the code expects there to be files like lib/gemstash/man/gemstash-readme.7. Pondering.

PR #401, opened 2024-11-19
Fix failing spec on ruby 3.3

segiddins:
Signed-off-by: Samuel Giddins <segiddins@segiddins.me>

# Description:

______________

# Tasks:

-  Describe the problem / feature
-  Write tests

I will abide by the [code of conduct](https://github.com/gemstash/gemstash/blob/master/CODE_OF_CONDUCT.md).

PR #407, opened 2025-03-28
Unskip and fix tests that fail depending on user's local .gemrc file.

martinemde:
# Description:

Without writing .gemrc, the gem source -a and -r don't work correctly. In particular, the presence of update_sources: true in .gemrc breaks one test, while writing the .gemrc fixes a previously skipped test.

______________

# Tasks:

- Fixed local tests for different .gemrc configurations.
- Un-skipped a test that works now with .gemrc written.

### DIFF
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -152,18 +152,27 @@ RSpec.describe "gemstash integration tests" do
       end
 
       it "finds private gems when just the private source is configured", db_transaction: false do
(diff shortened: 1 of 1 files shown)


martinemde:
Hey @segiddins, looks like you already took a stab at something like this in #401. This is passing and works for me locally on 3.1, 3.2, 3.3, and 3.4. The catch seems to be if you have `update_sources: true` in `~/.gemrc`.

martinemde (commented):
I'm going to merge this to fix tests for local dev. Please update if this causes any issues.

PR #408, opened 2025-04-04
Update Gemstsah::Logging.logger to use configured logfile

technicalpickles:
# Description:

This was hard-coded to $stdout, but should use what a user configures.
______________

# Tasks:

-  Write tests

### DIFF
--- a/lib/gemstash/logging.rb
+++ b/lib/gemstash/logging.rb
@@ -47,8 +47,6 @@ module Gemstash
     end
 
     def self.logger
-      require "debug"
-      debugger
       @logger ||= setup_logger(gemstash_env.log_file)
     end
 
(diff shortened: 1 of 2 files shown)


technicalpickles:
I just realized now that this is being used, gemstash ends up defaulting to the server.log. That means that there is no output when running `gemstash start`.

If that is the case, and the logfile hasn't been used, I think it'd make sense to default log_file to stdout instead

lib/gemstash/logging.rb
```diff
@@ -61,6 +63,7 @@ def initialize(app)
 
       def call(env)
         env["rack.logger"] = Gemstash::Logging.logger
+        env["rack.errors"] = Gemstash::Env.current.log_file
```
technicalpickles, line 66: I added this when I was getting errors down in rack when trying to use `rack.errors` to build a `rack.logger` object. I think that was actually having an exception when reporting an exception though.

I'll try reproducing to see how much this is needed

NathanZook:
@technicalpickles I'm seeing a bunch of failing tests.  What is the status here?

technicalpickles:
I don't recall the errors, but updating the branch from main was enough to fix it 🤷🏻

